PR #24, opened 2017-08-10
GetAll filter / keys-only

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Standard disclaimer about commits not being displayed in the sensible order :(

src/dhtproto/client/request/internal/GetAll.d
```diff
@@ -490,23 +492,45 @@ private scope class GetAllHandler
...
+                    this.outer.context, notification) )
                 {
                     // The user used the controller in the notifier callback
                     this.outer.resources.request_event_dispatcher.signal(
                         this.outer.conn,
                         suspendable_control.Signal.StateChangeRequested);
                 }
+            }
+
+            if ( this.outer.context.user_params.args.settings.keys_only )
```
nemanja-boric-sociomantic, line 507: Maybe it's better to do this with a single `foreach` loop (in both cases iterating over keys and values (if that's possible) and `if` inside its body. I'm not sure.

gavin-norman-sociomantic: It's possible but would be strange, I feel. It'd need to be something like:
```D
foreach ( e1, e2; batch )
{
  if ( this.outer.context.user_params.args.settings.keys_only )
  {
    send_key_to_user(e1);
    send_key_to_user(e2);
  }
  else
    send_record_to_user(e1, e2);
}
```

gavin-norman-sociomantic: (And that's only possible with the current way the batcher works, where the key and value are both stored as strings. This will likely change in the future.)

nemanja-boric-sociomantic: Fair enough

gavin-norman-sociomantic:
Rebased.

src/dhtproto/client/request/GetAll.d
```diff
@@ -24,6 +24,25 @@ public import dhtproto.client.NotifierTypes;
...
 /*******************************************************************************
 
+    GetAll behaviour settings. May be optionally passed to the getAll() method
+    of the DHT client, to modify the default behaviour.
+
+*******************************************************************************/
+
+public struct Settings
+{
+    import dhtproto.common.GetAll;
```
nemanja-boric-sociomantic: Why this import here?

src/dhttest/util/LocalStore.d
```diff
@@ -408,4 +409,61 @@ public struct NeoVerifier
...
+        }
+
+        DhtClient.Neo.GetAll.Settings settings;
+        settings.keys_only = true;
+        dht.neo.getAll(channel.dup, &notifier, settings);
+        task.suspend();
+
+        test(!duplicate);
+        test!("==")(keys.length, this.local.data.length);
+        foreach ( k, v; this.local.data )
```
nemanja-boric-sociomantic, line 467: I suggest to test `v.length == 0` as well.

src/dhtproto/node/neo/request/GetAll.d
```diff
@@ -56,6 +56,61 @@ public abstract scope class GetAllProtocol_v0
...
+
+            Returns:
+                true if the value passes (i.e. matches) the filter or if
+                filtering is not active
+
+        ***********************************************************************/
+
+        public bool match ( in void[] value )
+        {
+            if ( this.active )
```
nemanja-boric-sociomantic, line 104: `assert (this.matcher !is null)`

src/dhtproto/client/request/GetAll.d
gavin-norman-sociomantic: True, no longer needed.

src/dhttest/util/LocalStore.d
gavin-norman-sociomantic: Hmm maybe I'm misunderstanding what you mean, but I think that's irrelevant.

A test case puts some records in the DHT and the same records in the local store. A bunch of verification functions run, including this one, which only checks the keys returned by a keys-only GetAll vs the keys in the local store. Why are the values in the local store relevant?

nemanja-boric-sociomantic: I'm sorry, my bad. I read this as a `task-blocking` `GetAllKeys` request.

src/dhtproto/node/neo/request/GetAll.d
gavin-norman-sociomantic: I think this is impossible. `SearchFruct!(T)` is a struct and the internal "has this been setup?" fields are private.

nemanja-boric-sociomantic: :+1:

src/dhtproto/client/request/GetAll.d
gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

PR #26, opened 2017-08-28
Implement void[]-based ring queue in Mirror handler

gavin-norman-sociomantic:
Fixes two issues:
1. We encountered segfaults under high load. Investigation revealed that
   this is due to casting a `void[]*` to a pointer to another type of array
   (as we were doing in the old queue implementation). The exact cause of
   the segfaults was not discovered, but it appears to be related to GC
   corruption.
   See https://github.com/sociomantic-tsunami/dhtnode/issues/15.
2. The previous queue implementation was pretty inefficient: it required a
   memory shift of the entire queue after each pop. The new implementation
   does not require any memory shifting.

### DIFF
--- a/src/dhtproto/node/neo/request/Mirror.d
+++ b/src/dhtproto/node/neo/request/Mirror.d
@@ -14,6 +14,11 @@ module dhtproto.node.neo.request.Mirror;
 
 import ocean.util.log.Log;
 
+version ( UnitTest )
+{
+    import ocean.core.Test;
+}
+
 /// ditto
 public abstract scope class MirrorProtocol_v0
 {
@@ -53,8 +58,7 @@ public abstract scope class MirrorProtocol_v0
 
     /***************************************************************************
 
-        Struct template implementing a simple, array-based queue.
-        TODO use a proper ring queue to avoid mem shifting
+        Struct template implementing a simple, typed, array-based ring queue.
 
         Params:
             E = type of queue element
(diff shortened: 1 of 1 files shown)


src/dhtproto/node/neo/request/Mirror.d
```diff
@@ -260,7 +371,8 @@ public abstract scope class MirrorProtocol_v0
                 (the request code and version are assumed to be extracted)
 
     ***************************************************************************/
-
+import ocean.io.Stdout;
```
gavin-norman-sociomantic: Haha, ok I'll remove this debug stuff ;)

src/dhtproto/node/neo/request/Mirror.d
```diff
@@ -65,18 +69,50 @@ public abstract scope class MirrorProtocol_v0
...
+            Params:
+                buf = array to use to back queue storage
+
+        ***********************************************************************/
+
+        public void initialise ( void[]* buf )
+        {
+            this.max_elems = (max_size / E.sizeof) * E.sizeof;
+            buf.length = this.max_elems * E.sizeof;
+            this.queue = (cast(E*)buf.ptr)[0 .. this.max_elems];
```
david-eckardt-sociomantic: Suggest
- `ref void[] buf` now that you don't store a pointer to the array any more,
- `(cast(E[])*buf)[0 .. this.max_elems]` (without the `*` for `ref void[] buf`).

david-eckardt-sociomantic: Actually `this.queue = cast(E[])buf;` should be sufficient

gavin-norman-sociomantic: Updated.

david-eckardt-sociomantic (approved):

nemanja-boric-sociomantic (approved):

joseph-wakeling-sociomantic:
Can confirm that this (in combination with swarm changes in v4.2.4) results in no segfaults during a mirror request for my use-case.

PR #28, opened 2017-08-31
Add missing breaking change note re: GetAll notifications

gavin-norman-sociomantic:


nemanja-boric-sociomantic:
I think we're ready for the final `switch` with the new compilers, maybe we should update the `UsageExamples`.

nemanja-boric-sociomantic (approved):

PR #29, opened 2017-08-31
Switch to non-deprecated delegate type when calling sformat function

joseph-wakeling-sociomantic:
A couple of calls to `ocean.text.convert.Formatter.sformat` continue to use the deprecated `size_t delegate (cstring)` delegate format, perhaps in order to ensure compatibility with older ocean versions.

Since all supported minor releases of ocean (v3.2.x and v3.3.x) support the newer `void delegate (cstring)` (aka `FormatterSink`), and the `neo` branch is in any case pre-release code, it seems better to switch over to the new way of doing things.

### DIFF
--- a/src/dhtproto/client/NotifierTypes.d
+++ b/src/dhtproto/client/NotifierTypes.d
@@ -46,11 +46,7 @@ public struct RequestRecordInfo
     public void toString ( void delegate ( cstring chunk ) sink )
     {
         Formatter.sformat(
-            ( cstring chunk )
-            {
-                sink(chunk);
-                return chunk.length;
-            },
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic:
Thanks! I think this is probably a problem in swarm too.

nemanja-boric-sociomantic:
It's fine in swarm IIRC: https://github.com/sociomantic-tsunami/swarm/pull/74

gavin-norman-sociomantic:
:+1: Ah yes.

PR #32, opened 2017-09-07
Implement neo GetChannels request

gavin-norman-sociomantic:


src/dhtproto/client/request/internal/GetChannels.d
```diff
@@ -0,0 +1,367 @@
...
+            *this.conn.message_parser.getValue!(MessageType)(payload);
+        with ( MessageType ) switch ( msg_type )
+        {
+            case ChannelName:
+                Const!(void)[] channel;
+                this.conn.message_parser.parseBody(payload, channel);
+                Notification notification;
+                notification.received =
+                    RequestDataInfo(context.request_id, channel);
+                GetChannels.notify(this.context.user_params, notification);
```
nemanja-boric-sociomantic, line 352: Hm. Does this imply user receiving duplicate channel names in the notifier?

gavin-norman-sociomantic: Yes. This request has always worked like that.

Maybe we should think of something better, though. Hadn't occurred to me until you mentioned this :(

nemanja-boric-sociomantic: Yeah, I think it shouldn't be hard to have a set of channels that user already received notification for inside `SharedResources`.

gavin-norman-sociomantic: Will do. This will require two things:
1. A template in swarm for structs wrapping a `void[]` and treating it as an array of another type.
2. Acquiring two arrays. Append channel names to one and store slices to the first in the second.

gavin-norman-sociomantic: This is problematic at the moment. See https://github.com/sociomantic-tsunami/dhtproto/issues/34.

Suggest continuing to review / merge this PR as-is.

src/dhtproto/client/request/internal/GetChannels.d
```diff
@@ -0,0 +1,367 @@
...
+                break;
+
+            case Finished:
+                finished = true;
```
nemanja-boric-sociomantic: Here just `return true` and `return false` at the end of the method? Saves you a boolean variable. Maybe current approach is cleaner.

src/dhtproto/node/neo/request/GetChannels.d
```diff
@@ -1 +1,89 @@
...
+            }
+        );
+
+        auto channel_buf = this.resources.getVoidBuffer();
+
+        foreach ( channel; this )
+        {
+            // Copy the channel name in case it changes during sending.
+            (*channel_buf).copy(channel);
+            enableStomping(*channel_buf);
```
nemanja-boric-sociomantic: I think `copy` already does this.

src/dhtproto/node/neo/request/GetChannels.d
```diff
@@ -1 +1,89 @@
...
+        );
+    }
+
+    /***************************************************************************
+
+        opApply iteration over the names of the channels in storage.
+
+    ***************************************************************************/
+
+    protected abstract int opApply ( int delegate ( ref cstring ) dg );
```
nemanja-boric-sociomantic, line 87: Nitpick: this is a bit odd. I find `opApply` to be a member of the struct we're iterating over, but here it's the iteration over some other entity. I can't think of any better (which is also as easy to implement - obviously, you could declare a base abstract class here, and make an abstract method returning an instance of it, but that's just too much work for very little benefit) solution.

src/fakedht/neo/request/GetChannels.d
```diff
@@ -0,0 +1,111 @@
...
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module fakedht.neo.request.GetChannels;
+
+/*******************************************************************************
```
nemanja-boric-sociomantic: Please no `Imports` :D

gavin-norman-sociomantic: Haha, just cut & pasted from another module ;)

src/dhtproto/client/mixins/NeoSupport.d
```diff
@@ -1097,6 +1097,165 @@ template NeoSupport ( )
...
+                with ( info.Active ) switch ( info.active )
+                {
+                    case received:
+                        // Ignore all received value on user break.
+                        if (this.state == State.Stopped)
+                            break;
+
+                        copy(*this.channel_name,
+                            cast(cstring)info.received.value);
+                        enableStomping(*this.channel_name);
```
nemanja-boric-sociomantic: Ditto: I think `copy` does this already.

nemanja-boric-sociomantic:
Few small comments, in general LGTM

src/dhtproto/client/mixins/NeoSupport.d
gavin-norman-sociomantic: At least in ocean v3.3.1 (used by dhtproto) it doesn't. https://github.com/sociomantic-tsunami/ocean/blob/v3.3.1/src/ocean/core/array/Mutation.d#L925-L933

(Ditto for v3.x.x.)

nemanja-boric-sociomantic: The first argument here is `Buffer`, and `dest.length` does that internally: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/core/Buffer.d#L140

src/dhtproto/node/neo/request/GetChannels.d
gavin-norman-sociomantic: I know what you mean. But yeah, the reason I did it like this is for convenience / simplicity.

src/dhtproto/client/mixins/NeoSupport.d
gavin-norman-sociomantic: Ah indeed.

src/dhtproto/client/request/internal/GetChannels.d
gavin-norman-sociomantic: No I think a direct `return` is cleaner.

src/dhtproto/node/neo/request/GetChannels.d
nemanja-boric-sociomantic: I know what you mean.

gavin-norman-sociomantic:
Updated.

nemanja-boric-sociomantic (approved):

PR #33, opened 2017-09-07
Fix log Message for connetion_io-stats

stefan-koch-sociomantic:
Just adding a parameter to the log-string where it was missing


stefan-koch-sociomantic:
@gavin-norman-sociomantic pr created.

gavin-norman-sociomantic:
Thanks!

PR #35, opened 2017-09-18
Extract logic for get/put node selection

gavin-norman-sociomantic:
(Based on #32.)

### DIFF
--- a/src/dhtproto/client/internal/NodeHashRanges.d
+++ b/src/dhtproto/client/internal/NodeHashRanges.d
@@ -118,6 +118,7 @@ private class NodeHashRangesBase
     import swarm.neo.AddrPort;
     import swarm.neo.client.ConnectionSet;
     import swarm.util.Hash : isWithinNodeResponsibility;
+    import ocean.core.array.Mutation : sort;
 
     /// Value of the next created NodeHashRange's order field.
     private static ulong order_counter;
@@ -170,7 +171,9 @@ private class NodeHashRangesBase
     /***************************************************************************
 
         Gets the list of nodes (along with hash range and ordering information)
-        which cover the specified hash.
+        which cover the specified hash, sorted (descending) by the values of
+        their `order` fields. (This means that the node most recently reported
+        as covering the specified hash will appear first in the list.)
 
         Params:
             h = hash to query
@@ -198,6 +201,14 @@ private class NodeHashRangesBase
             }
         }
 
+        bool sortPred ( NodeHashRange e1, NodeHashRange e2 )
+        {
+            assert(e1.order != e2.order);
+            return e1.order > e2.order;
+        }
+
+        node_hash_ranges.sort(&sortPred);
+
         return node_hash_ranges;
     }
 
@@ -223,29 +234,16 @@ private class NodeHashRangesBase
 version ( UnitTest )
 {
     import ocean.core.Test;
-    import ocean.core.array.Mutation : sort;
     import Integer = ocean.text.convert.Integer_tango;
     import swarm.neo.AddrPort;
     import ocean.core.BitManip : bitswap;
-}
 
-unittest
-{
     alias NodeHashRange.HashRange HR;
 
     void checkTestCase ( NodeHashRange[] r1, NodeHashRange[] r2, long line_num = __LINE__ )
     {
         auto t = new NamedTest(idup("Test at line " ~ Integer.toString(line_num)));
         t.test!("==")(r1.length, r2.length);
-
-        bool sortPred ( NodeHashRange e1, NodeHashRange e2 )
-        {
-            return e1.addr.cmp_id < e2.addr.cmp_id;
-        }
-
-        // Ordering doesn't matter, so we sort both arrays
-        r1.sort(&sortPred);
-        r2.sort(&sortPred);
 
         foreach ( i, e; r1 )
         {
@@ -255,7 +253,11 @@ unittest
             t.test!("==")(e.order, r2[i].order);
         }
     }
+}
 
+// Tests for hash range overlaps and gaps
+unittest
+{
     auto addr1 = AddrPort(1, 1);
     auto addr2 = AddrPort(2, 2);
 
@@ -279,8 +281,8 @@ unittest
         [NodeHashRange(addr1, HR(hash_t.min, hash_t.max), 0)]);
     hr.getNodesForHash(hash_t.max, ranges);
     checkTestCase(ranges,
-        [NodeHashRange(addr1, HR(hash_t.min, hash_t.max), 0),
-         NodeHashRange(addr2, HR(0x8000000000000000, hash_t.max), 1)]);
+        [NodeHashRange(addr2, HR(0x8000000000000000, hash_t.max), 1),
+         NodeHashRange(addr1, HR(hash_t.min, hash_t.max), 0)]);
 
     // Change the range of the first node to cover the other half of the range
     hr.updateNodeHashRange(addr1, hash_t.min, 0x7fffffffffffffff);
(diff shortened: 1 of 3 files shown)


src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -57,6 +59,115 @@ public final class NodeHashRanges : NodeHashRangesBase
...
+            During a data redistribution, more than one node may be responsible
+            for a given key. In this case, the node that was most recently
+            reported as being responsible for the key is queried first, followed
+            by others (in order) until the record is located, an error occurs,
+            or no node has the record.
+
+        TODO: test the logic for retrying the request on other nodes which
+        previously covered the hash. This will require a full neo implementation
+        of the Redistribute request. See
+        https://github.com/sociomantic/dhtnode/issues/624
```
nemanja-boric-sociomantic: non-tsunami link.

src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -57,6 +59,115 @@ public final class NodeHashRanges : NodeHashRangesBase
...
+
+    ***************************************************************************/
+
+    public bool getFromNode ( hash_t h, ref NodeHashRange[] node_hash_ranges,
+        UseNodeDg use_node,
+        bool delegate ( RequestOnConn.EventDispatcher ) get )
+    {
+        auto nodes = this.getNodesForHash(h, node_hash_ranges);
+        if ( nodes.length == 0 )
+            return true;
```
nemanja-boric-sociomantic: Why not make this return code inverse? If there's no record, return `false`, that seems more logical.

src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -57,6 +59,115 @@ public final class NodeHashRanges : NodeHashRangesBase
...
+
+        foreach_reverse ( node_hash_range; nodes )
```
nemanja-boric-sociomantic: This `foreach_reverse` indicates that the ordering is somehow important. Maybe it should be added to the method's documentation for the parameter `getNodesForHash`.

src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -57,6 +59,115 @@ public final class NodeHashRanges : NodeHashRangesBase
...
+                    try_next_node = get(conn);
+                }
+            );
+
+            // If we got the record or an error occurred, don't try more nodes
+            if ( !try_next_node )
+                break;
+        }
+
+        return false;
```
nemanja-boric-sociomantic: This fixed `return false` doesn't seem to be covering the case where we get the list of nodes responsible for the key, but they all return `true`. Perhaps rewrite this as:

```
+            // If we got the record or an error occurred, don't try more nodes
 +            if ( !try_next_node )
 +                return true;
 +        }
 +
 +        return false;
```

nemanja-boric-sociomantic: This fixed `return false` seems to ignore the case where we tried all nodes that we got as the responsible, but nobody contains the record anymore. Is that plausible scenario to keep yourself from?

src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -57,6 +59,115 @@ public final class NodeHashRanges : NodeHashRangesBase
...
+    {
+        auto nodes = this.getNodesForHash(h, node_hash_ranges);
+        if ( nodes.length == 0 )
+            return true;
+
+        size_t newest;
+        ulong newest_order;
+        foreach ( i, node_hash_range; nodes )
+        {
+            if ( node_hash_range.order > newest_order )
```
nemanja-boric-sociomantic: To confirm: `node_hash_range.order` starts from `1`? Maybe it's safer to do `if (node_hash_range.order >= newest_order)`.

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: It already is described...?

>During a data redistribution, more than one node may be responsible
for a given key. In this case, the node that was most recently
reported as being responsible for the key is queried first, followed
by others (in order) until the record is located, an error occurs,
or no node has the record.

gavin-norman-sociomantic: Oh I see what you mean. `getNodesForHash` doesn't mention how the returned list is ordered.

gavin-norman-sociomantic: Weird, it appears it's actually unsorted! (I think this is what the TODO above is referring to ;)

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: :+1: I'll create a new issue in tsunami.

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: I know what you mean. The most common usage requires a `!` then, but I guess that doesn't really matter.

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: There should never be two nodes with the same `order`. I can add checks that verify this.

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: The caller can already detect this case (they are the ones who decide whether on the return value of `get`), so it doesn't need to be specifically covered here.

gavin-norman-sociomantic: The caller can already detect this case (they are the ones who decide whether on the return value of get), so it doesn't need to be specifically covered here.

gavin-norman-sociomantic: Actually, the return value isn't needed at all. The caller can detect all possible cases externally.

src/dhtproto/client/internal/NodeHashRanges.d
nemanja-boric-sociomantic: No, it was my brain that stuck yesterday. This piece of code is actually fine.

gavin-norman-sociomantic:
Updated.

src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -198,6 +201,14 @@ private class NodeHashRangesBase
             }
         }
 
+        bool sortPred ( NodeHashRange e1, NodeHashRange e2 )
+        {
+            assert(e1.order != e2.order);
```
nemanja-boric-sociomantic, line 296: Too early for `verify`?

src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -57,6 +59,94 @@ public final class NodeHashRanges : NodeHashRangesBase
...
+                nodes which cover the specified hash
+            use_node = delegate to call to gain access to a request-on-conn to
+                communicate with the selected node
+            get = delegate to communicate with the selected node (over the
+                request-on-conn provided by use_node). Should return true if the
+                get succeeded or false to try the next node
+
+    ***************************************************************************/
+
+    public void getFromNode ( hash_t h, ref NodeHashRange[] node_hash_ranges,
```
nemanja-boric-sociomantic, line 88: Yeah, without return value what this method is doing (and how it should be used) is much clearer.

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: Mmm... Is it released yet?

nemanja-boric-sociomantic: Yes, in ocean `v2.10.0`/`v3.4.0`.

gavin-norman-sociomantic: Cool, I can update.

gavin-norman-sociomantic:
Rebased on `neo`. Unblocked.

src/dhtproto/client/internal/NodeHashRanges.d
nemanja-boric-sociomantic: > Cool, I can update.

I assume you figured out that you can't update, since we don't support this ocean yet?

nemanja-boric-sociomantic (approved):

src/dhtproto/client/internal/NodeHashRanges.d
gavin-norman-sociomantic: Oh I forgot about this. What do you mean we don't support it yet?

nemanja-boric-sociomantic: If dhtproto/swarm is using ocean prior to `v3.4.0`, `verify` is not yet available.

gavin-norman-sociomantic: I could update the submodule in this (semver-free) branch.

gavin-norman-sociomantic: I won't bother. It introduces zillions of deprecation warnings. Something for a separate PR.

PR #42, opened 2017-09-29
Update to swarm v4.3.0 and remove deprecations

gavin-norman-sociomantic:

PR #43, opened 2017-10-02
Reject big records

gavin-norman-sociomantic:
Based on #42. (Blocked.)

### DIFF
--- a/relnotes/record-size-limit.breaking.md
+++ b/relnotes/record-size-limit.breaking.md
@@ -1,0 +1,10 @@
+* `dhtproto.client.request.Put`
+
+  The neo Put request now enforces a size limit on record values written to the
+  DHT. Any record value which is larger than the constant defined in this module
+  will be rejected. The user will be notified of this via the new
+  `value_too_big` notifier.
+
+  Put notifiers in existing code should be updated to handle the new
+  notification type as appropriate.
+
--- a/src/dhtproto/client/UsageExamples.d
+++ b/src/dhtproto/client/UsageExamples.d
@@ -219,6 +219,7 @@ unittest
                     break;
 
                 case failure:
+                case value_too_big:
                 case no_node:
                 case node_disconnected:
                 case node_error:
@@ -262,6 +263,7 @@ unittest
                     break;
 
                 case failure:
+                case value_too_big:
                 case no_node:
                 case node_disconnected:
                 case node_error:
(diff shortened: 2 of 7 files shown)


david-eckardt-sociomantic:
What would happen currently, without checking the record size and responding with `ValueTooBig`, if a batch request comes across a too big record? Would the batch request just fail with `Error`?

david-eckardt-sociomantic:
Can you even add a record with the length of `RecordBatcher.DefaultMaxBatchSize` to a batch? I mean, first the `size_t` length of the record will be added to the batch, then the record data, so the batch length should be `size_t.sizeof + DefaultMaxBatchSize` then -- perhaps `DefaultMaxBatchSize - size_t.sizeof` is the actual record size limit?

gavin-norman-sociomantic:
>What would happen currently, without checking the record size and responding with ValueTooBig, if a batch request comes across a too big record? Would the batch request just fail with Error?

Currently, the GetAll request just skips the too big record, logs a warning, and continues with the iteration: https://github.com/sociomantic-tsunami/dhtproto/blob/neo/src/dhtproto/node/neo/request/GetAll.d#L314-L319

gavin-norman-sociomantic:
>Can you even add a record with the length of RecordBatcher.DefaultMaxBatchSize to a batch? I mean, first the size_t length of the record will be added to the batch, then the record data, so the batch length should be size_t.sizeof + DefaultMaxBatchSize then -- perhaps DefaultMaxBatchSize - size_t.sizeof is the actual record size limit?

Good point. I think you're probably right. I'l investigate this and add a test with GetAll to ensure that a record with the maximum allowed value size can indeed be successfully returned to the client via GetAll.

(I believe the legacy protocol has the same problem, btw.)

gavin-norman-sociomantic:
Actually, the best way to phrase this would be for the DHT's batch size to be set in terms of `MaxRecordSize + size_t.sizeof`. The batch size is an internal detail of the GetAll request, so it doesn't make sense to require the Put request to know about this.

gavin-norman-sociomantic:
Updated:
* One record of max size, including key, will fit in a GetAll batch.
* Add a test that a record of max size can be returned by GetAll.

gavin-norman-sociomantic:
#42 merged. Unblocked.

david-eckardt-sociomantic (approved):

PR #49, opened 2017-10-16
Merge v13.x.x -> neo

gavin-norman-sociomantic:


.travis.yml
```diff
@@ -19,12 +19,14 @@ env:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     matrix:
-        - DMD=1.081.1 DIST=xenial F=production AFTER_SCRIPT=1
-        - DMD=1.081.1 DIST=xenial F=devel
-        - DMD=2.070.2.s13 DIST=xenial F=production
-        - DMD=2.070.2.s13 DIST=xenial F=devel
+        - DMD=1.081.* DIST=xenial F=production AFTER_SCRIPT=1
```
gavin-norman-sociomantic, line 22: @nemanja-boric-sociomantic could you confirm this merge?

nemanja-boric-sociomantic: I confirm this is right. This will make apt always install the latest patch release of the tools without any developer's intervention, which is what we want.

gavin-norman-sociomantic: :+1: thanks. I wasn't sure why it was set to specific versions before.

nemanja-boric-sociomantic: We didn't know it was possible to do this :-). We discovered it when we released last patch releases for the compilers and went to update all `.travis.yml` files.

PR #55, opened 2017-10-27
Change neo GetAll.getNext to work with a delegate, not a buffer copy

gavin-norman-sociomantic:
This allows the storage engine to pass a slice of its contents to the
request handler's delegate, rather than requiring a copy into the handler's
buffer for each record. This is an optimisation that reduces CPU load
during iteration over large channels.

Part of https://github.com/sociomantic-tsunami/dhtnode/issues/19

### DIFF
--- a/src/dhtproto/node/neo/request/GetAll.d
+++ b/src/dhtproto/node/neo/request/GetAll.d
@@ -29,9 +29,6 @@ public abstract scope class GetAllProtocol_v0
     /// Mixin the constructor and resources member.
     mixin RequestCore!();
 
-    /// Buffer used to store record values to be sent to the client.
-    private void[]* value_buffer;
-
     /// Batch of records to send.
     private RecordBatcher batcher;
 
@@ -161,7 +158,6 @@ public abstract scope class GetAllProtocol_v0
         this.filter.init(value_filter);
 
         // Acquire required resources.
-        this.value_buffer = this.resources.getVoidBuffer();
         this.compressed_batch = this.resources.getVoidBuffer();
         this.batcher = this.resources.getRecordBatcher();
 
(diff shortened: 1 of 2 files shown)


src/dhtproto/node/neo/request/GetAll.d
```diff
@@ -286,40 +282,46 @@ public abstract scope class GetAllProtocol_v0
             uint yield_counter;
 
             // Iterate over the channel and send each record to the client.
-            while ( this.outer.getNext(key, *this.outer.value_buffer) )
+            bool more = true;
+            while ( more )
```
nemanja-boric-sociomantic: `bool more; do { ... } while (more);`?

src/dhtproto/node/neo/request/GetAll.d
```diff
@@ -231,16 +227,16 @@ public abstract scope class GetAllProtocol_v0
...
-            true if a record was returned via the out arguments or false if the
-            iteration is finished
+            true if a record was passed to `dg` or false if the iteration is
+            finished
 
     ***************************************************************************/
 
-    abstract protected bool getNext ( out hash_t key, ref void[] value );
+    abstract protected bool getNext (
+        void delegate ( hash_t key, Const!(void)[] value ) dg );
```
nemanja-boric-sociomantic, line 239: I wonder if this delegate should return a boolean indicating if it want's to continue the operation, akin to `opApply`, but I can't think of the compelling reason right now, as unlike `opApply`, user is responsible of calling `getNext` for each next record.

nemanja-boric-sociomantic: (in which case, returning false from the delegate would make getNext return false as well).

gavin-norman-sociomantic: This delegate's function is just to fetch the current record; it has nothing to do with the decision about whether to continue or not.

nemanja-boric-sociomantic: That's true right now, I'm just wondering if it makes sense to allow it to influence decision if the iteration should continue or not. Probably not, I wouldn't complicate it without any reason :+1:.

src/dhtproto/node/neo/request/GetAll.d
gavin-norman-sociomantic: Updated.

src/fakedht/neo/request/GetAll.d
```diff
@@ -167,27 +167,29 @@ private scope class GetAllImpl_v0 : GetAllProtocol_v0
...
             return false;
 
         auto str_key = this.iterate_keys[$-1];
         this.iterate_keys.length = this.iterate_keys.length - 1;
 
+        hash_t key;
         auto ok = toHashT(str_key, key);
         assert(ok);
-        value = this.channel.get(key).dup;
+        auto value = this.channel.get(key);
```
nemanja-boric-sociomantic, line 190: Look, ma', no `dup`!

nemanja-boric-sociomantic (approved):

PR #57, opened 2017-11-01
Use SchedulingDhtClient.Config in neo SchedulingDhtClient constructor

joseph-wakeling-sociomantic:
This patch fixes an apparent oversight in providing the neo constructor for `SchedulingDhtClient`: there is no need for a separate `max_events` parameter, given that it is already provided in the class' own custom `Config` class.

This should make it easier for users to transition to the neo protocol as it maintains consistency with the other `Config`-using constructor overloads of `SchedulingDhtClient`.


gavin-norman-sociomantic (approved):
Thanks!

Issue #39, opened 2017-09-28
Prohibit RemoveChannel from deleting a channel being listened to


gavin-norman-sociomantic:
#10 discusses problems around ending a Mirror request when the channel being mirrored is deleted. The same argument applies to the legacy Listen request.

Discussion led to the idea that it should actually be illegal to remove a channel that is being listened to. RemoveChannel is used in two cases:
1. By accident.
2. Deliberately, to remove a channel which is no longer in use.

In both cases, the request should definitely not work on a channel that's currently in use.

joseph-wakeling-sociomantic:
Should this constraint apply only for channels being listened to, or for channels with any open request of any kind (the remove-channel request aside)?

gavin-norman-sociomantic:
If a request for a non-existent channel is handled, the node will create the channel. For this reason, other requests aren't really affected by this issue.

joseph-wakeling-sociomantic:
I was thinking of a `GetAll` request where the channel was removed part-way through.

gavin-norman-sociomantic:
The request would just be cut short as all data would be removed from the storage engine.

joseph-wakeling-sociomantic:
OK.  As long as that's an acceptable outcome from your point of view, I have no objections.  I wonder if it might be worth checking in with @scott-gibson-sociomantic though, since it's his use case that `GetAll` is there to support.

gavin-norman-sociomantic:
We're talking about the legacy requests (including GetAll) as well.

joseph-wakeling-sociomantic:
Yes, understood.  But since there is a neo `GetAll` I think it's worth checking in on the use-case before assuming that it's OK for it to just cut out part-way through on channel removal.

scott-gibson-sociomantic:
Correct me if i'm wrong but doesn't neo have a `Deleted` notification in the notifier to indicate the channel was deleted? It's been a while since i looked so my memory is hazy.

joseph-wakeling-sociomantic:
@scott-gibson-sociomantic well, we're discussing whether channel removal should be rejected if there is at least one listener still in place.  The question is whether your own neo `GetAll` use-case would also find it problematic if the channel was deleted part-way through.

gavin-norman-sociomantic:
[...]

PR #61, opened 2017-11-09
Modify legacy RemoveChannel behaviour to not remove listened channels

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Thinking about release notes for this:
* It's not a breaking change to the API.
* It's also not a breaking change to the protocol.
* It _is_, however, a change to the request's behaviour.

We don't really have a release notes section for such changes, at present. Maybe we could add a new category for the *proto repos? (Though that would probably require support in `neptune-release`.) Or we just count it as a migration instruction? Any other ideas?

nemanja-boric-sociomantic:
I would keep it simple and count this as a migration instructions, as I'm inclined to look into this as a _behavioral_ breaking change, as the legitimate use cases that used to work now doesn't work anymore (for example, there are legitimate cases where we are resetting channels in the DHT by removing the channel, which will not work anymore, unless all listeners are down, we're not just using RemoveChannel to remove the orphan channels).

gavin-norman-sociomantic:
>there are legitimate cases where we are resetting channels in the DHT by removing the channel

Ah, interesting. I'd forgotten about those. Essentially using `RemoveChannel` as a "delete all records". Ah well, such uses are obscure.

gavin-norman-sociomantic:
Updated to the Nemanja solution: each test case operates on its own channel.

gavin-norman-sociomantic:
Updated with release notes.

gavin-norman-sociomantic:
This change is getting too complicated:
* As seen in this PR, changing the behaviour of RemoveChannel affects the legacy tests.
* Switching all test cases (in the neo branch) to use the neo RemoveChannel doesn't work, because of the different behaviour of Listen.
* We could implement a unix socket command to clear the state of the DHT after a test case, but then we wouldn't want to deploy this to live DHTs!

The impetus for this change was to simplify the Mirror request (see #39). While this would be a nice thing to do, it's far from urgent. Mirror works as it is, it just has a slight and obscure complication in its usage (user code must handle what to do if the mirrored channel is removed). Given the complications above, it's not worth making this change now. We can wait until the legacy requests are no longer in use, then make a new version of Mirror and RemoveChannel with this altered behaviour.

PR #81, opened 2017-12-19
Automatic D2 tag release

gavin-norman-sociomantic:
Copied from swarm.


mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Luca has already configured env vars for all proto repos so it should be good to go.

PR #90, opened 2018-01-08
Adapt Get, Put, GetChannels to use new SupportedCode from swarm

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Gah, it's too early in the year for flaky tests :(

gavin-norman-sociomantic:
Interesting... it seems that reading and checking the started / error code in the `handleSupportedCode` method (instead of in the `handle` method) fixes the flaky tests. I'm not entirely sure why.

gavin-norman-sociomantic:
Green now.

>Interesting... it seems that reading and checking the started / error code in the handleSupportedCode method (instead of in the handle method) fixes the flaky tests. I'm not entirely sure why.

@nemanja-boric-sociomantic interested whether you have any thoughts on that. The most obvious thing I can think of is what will happen in [`AllNodesRequest.run`](https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/client/mixins/AllNodesRequestCore.d#L127-L164). I don't see what difference that would make, though.

nemanja-boric-sociomantic:
I'm sceptical towards the fact that this caused/fixed flaky tests :thinking: .

gavin-norman-sociomantic:
Ditto.

gavin-norman-sociomantic:
It was very repeatable, though. With the previous change: ~2/4 builds would fail. With the current change: no builds fail.

gavin-norman-sociomantic:
How ironic that as soon as I type that, I notice that the current build has failed with the same problem :D

gavin-norman-sociomantic:
Moved flaky part to https://github.com/sociomantic-tsunami/dhtproto/pull/93.

nemanja-boric-sociomantic:
:+1: Looks very noncontroversal.

Issue #78, opened 2017-12-18
Adapt requests to use new request handler framework from swarm


gavin-norman-sociomantic:
Swarm v4.5.0 includes [a PR to convert the global status codes into a supported/not supported code](https://github.com/sociomantic-tsunami/swarm/pull/209).

All requests in dhtproto neo should be updated to use this system, rather than the old approach of an unsupported / ok / error status code. (The idea is that the initial status code _purely_ indicates whether the node supports the request or not, and all actual errors are handled by messages that can be passed at any time, not just at request initialisation.)

* [x] GetHashRange
* [x] Put
* [x] Get
* [x] GetAll
* [x] GetChannels
* [x] Mirror (#94)
* [x] Exists
* [x] Remove
* [x] RemoveChannel

gavin-norman-sociomantic:
One commit in the PR gives an example of adapting a request: https://github.com/sociomantic-tsunami/swarm/pull/209/commits/96a2c389135c2c6a91b292bd6de29b523cd9289e

gavin-norman-sociomantic:
This is a protocol change. Existing requests could be adapted via a new request version, but as we're in alpha still, backwards compatibility is of low importance. I'll simply do this as a breaking change.

gavin-norman-sociomantic:
https://github.com/sociomantic-tsunami/dhtproto/pull/90 adapts Get, Put, GetChannels, Mirror.

I'll do GetHashRange and GetAll as part of #70 and #38.

gavin-norman-sociomantic:
All done except GetAll. Waiting until https://github.com/sociomantic-tsunami/swarm/issues/243 is implemented.

gavin-norman-sociomantic:
GetAll was actually done in https://github.com/sociomantic-tsunami/dhtproto/pull/122/commits/aacd1fa0c6fa11bcc2fe546bfd6d9638074332d5, as part of the major submodule update.

PR #93, opened 2018-01-09
Adapt Mirror to use new SupportedCode from swarm

gavin-norman-sociomantic:
Part of #78.

### DIFF
--- a/src/dhtproto/client/request/internal/Mirror.d
+++ b/src/dhtproto/client/request/internal/Mirror.d
@@ -299,6 +299,18 @@ private scope class MirrorHandler
             return false; // Request/version not supported
         }
 
+        return true;
+    }
+
+    /***************************************************************************
+
+        Handler policy, called from AllNodesRequest template to run the
+        request's main handling logic.
+
+    ***************************************************************************/
+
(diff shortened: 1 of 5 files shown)


nemanja-boric-sociomantic:
This just got retested, let's see how flaky you are.

gavin-norman-sociomantic:
>Interesting... it seems that reading and checking the started / error code in the `handleSupportedCode` method (instead of in the `handle` method) fixes the flaky tests. I'm not entirely sure why.

Making this change seems to trigger flakiness in the Mirror tests. It's not clear, however, whether they were already flaky, and this change just ups the chances.

gavin-norman-sociomantic:
>This just got retested, let's see how flaky you are.

Flaky!

gavin-norman-sociomantic:
>It's not clear, however, whether they were already flaky, and this change just ups the chances.

Seems it's actually unrelated. We just had a failure in a build without this change: https://travis-ci.org/sociomantic-tsunami/dhtproto/jobs/327184612#L774

gavin-norman-sociomantic:
Replaced by #94.

PR #99, opened 2018-01-24
[DO NOT MERGE] Submodule update for v14.x.x

gavin-norman-sociomantic:
This PR contains all changes from the current v14.x.x, rebased onto v13.2.0 and updated to use the new release candidates of required submodules. It should not be merged into v13.x.x, but will be force pushed into v14.x.x.


integrationtest/dhtrestart/main.d
```diff
@@ -61,7 +62,7 @@ class RestartWithListeners : TestCase
 {
     import turtle.env.Dht;
 
-    import ocean.core.Test;
+    import Ocean = ocean.core.Test;
```
mihails-strasuns-sociomantic: Why is it necessary after renaming `test` package to `integrationtest`?

mihails-strasuns-sociomantic (approved):

integrationtest/dhtrestart/main.d
gavin-norman-sociomantic: Hm good question. I'll investigate.

gavin-norman-sociomantic: Ahh, that  commit comes _before_ the makd update.

gavin-norman-sociomantic: Reordered commits and removed the named import change.

gavin-norman-sociomantic:
Pushed to v14.x.x

PR #105, opened 2018-01-30
Pass request parameters to notifier as const

gavin-norman-sociomantic:


relnotes/args.migration.md
```diff
@@ -0,0 +1,6 @@
+# Request arguments now passed to notifiers as const
+
+`TODO`
```
david-eckardt-sociomantic: TODO ;)

gavin-norman-sociomantic:
Updated.

PR #94, opened 2018-01-11
Use new request handlers framework from swarm

gavin-norman-sociomantic:
(Requires swarm v4.6.0.)

Issue #16, opened 2017-08-08
neo Update request


gavin-norman-sociomantic:
Many applications need to update records in the DHT by:
1. Get the record's current value,
2. modify it in memory,
3. Put the modified version back to the DHT (overwriting the original value of the record).

The problem is that, if you have multiple instances of such an application running, race conditions can occur when 2 or more are updating the same record at the same time. This has been observed in the real world  and is a known and currently unfixable issue.

It should be possible to keep track of multiple concurrent Update requests on the same record and delay subsequent requests until the initial update has finished:

* When a client starts an Update request, the node allocates an object indicating that the record is locked.
* If another Update starts on the same record, the client is immediately informed that the record is locked.
* The client waits for a short time (10ms?) then restarts the request. (It is important that the client really starts again from the beginning, including re-fetching the list of nodes which are responsible for the record, which could have changed in the meantime, like in the race condition example above.)

Possible usage example, from the client's point of view:
```D
void notifier ( DhtClient.Neo.Update.Notification info, DhtClient.Neo.Update.Args args )
{
  with ( info.Active ) switch ( info.active )
  {
    // may be called multiple times, if other requests change the value
    case received:
      received.update(
        (in void[] record)
        {
          // TODO: update `record` as required
          return updated_record;
        }
      );
    break;

    // TODO: handle other cases...
  }
}

dht.neo.update(channel, key, &notifier);
```

gavin-norman-sociomantic:
A thought about writing a test for this:
1. Start an Update request.
2. When the "here's the record, give me your updated version" notification occurs, spawn a Task which starts a second Update request (for the same record) and then waits (== yield to epoll).
3. The second Update's notification should kick in while the first one's is still suspended, allowing us to test the behaviour of trying to update a record when someone else is already doing so.

gavin-norman-sociomantic:
We'll have to take care that the reading and writing logic works with the carefully designed behaviour of Put and Get to handle online data redistribution.

gavin-norman-sociomantic:
Summary of Put/Get redistribution handling logic:
1. Put requests always write to the newest known node which covers the record hash. (https://github.com/sociomantic-tsunami/dhtproto/blob/neo/src/dhtproto/client/request/internal/Put.d#L151-L160)
2. Get requests try nodes in newest -> oldest order. (https://github.com/sociomantic-tsunami/dhtproto/blob/neo/src/dhtproto/client/request/internal/Get.d#L161-L179)

[...]

PR #110, opened 2018-02-01
Update request

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Based on #94.

gavin-norman-sociomantic:
Initial sketch of client-side handling, for early browsing.

gavin-norman-sociomantic:
Added rough node-side implementation.

gavin-norman-sociomantic:
TODO: Try out https://github.com/sociomantic-tsunami/dhtproto/issues/16#issuecomment-364053337.

gavin-norman-sociomantic:
Updated: compiles and fully documented. Next step: testing.

gavin-norman-sociomantic:
>Updated: compiles and fully documented.

Well... compiles in D1 ;) That's good enough for now.

gavin-norman-sociomantic:
Testing under way, using https://github.com/sociomantic-tsunami/dhtnode/pull/90.

gavin-norman-sociomantic:
Local tests confirm the following:
* The basic Update procedure works.
* Update detects and rejects conflicts (where the record being updated changes while the request is in progress).

gavin-norman-sociomantic:
Updated, unblocked, ready for review.

src/dhtproto/client/UsageExamples.d
```diff
@@ -878,6 +878,119 @@ unittest
...
+            formatNotification(info, this.msg_buf);
+
+            with ( info.Active ) final switch ( info.active )
+            {
+                case received:
+                    // Set updated value. In this simple example, we simply
+                    // copy the original value, without change.
+                    auto received_record = info.received.value;
+                    (*info.received.updated_value).copy(info.received.value);
+                    goto case;
```
joseph-wakeling-sociomantic, line 911: Can you explain what this `goto` will do, and why?

gavin-norman-sociomantic: Fall-through to the next case (required by D2).

gavin-norman-sociomantic: Why? So that `log.trace` is called for this notification.

joseph-wakeling-sociomantic: Ah, gotcha.  I don't think I'd ever come across that syntax before, good to know.

gavin-norman-sociomantic: Yeah, it's kinda weird. Implicit fall-through is disallowed in D2.

gavin-norman-sociomantic:
Added a couple of simple tests.

gavin-norman-sociomantic:
...and a release note.

gavin-norman-sociomantic:
test this please dmd 2.071

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,329 @@
...
+        all I/O operations on the connection.)
+
+        Params:
+            init_payload = initial message payload read from the connection
+
+    ***************************************************************************/
+
+    public void preSupportedCodeSent ( Const!(void)[] init_payload )
+    {
+        auto ed = this.connection.event_dispatcher();
```
david-eckardt-sociomantic: Not needed.

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,329 @@
...
+    ***************************************************************************/
+
+    public void postSupportedCodeSent ( )
+    {
+        auto ed = this.connection.event_dispatcher();
+
+        auto message =
+            *ed.message_parser.getValue!(MessageType)(this.init_payload);
+        auto channel = ed.message_parser.getArray!(char)(this.init_payload);
+        auto key = *ed.message_parser.getValue!(hash_t)(this.init_payload);
```
david-eckardt-sociomantic, line 80: Suggest using `ed.message_parser.parseBody`.

david-eckardt-sociomantic: Never mind, you are using the tail of `this.init_payload` below.

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,329 @@
...
+            {
+                exists = true;
+
+                ed.send(
+                    ( ed.Payload payload )
+                    {
+                        payload.addCopy(MessageType.RecordValue);
+                    }
+                );
+                ed.flush();
```
david-eckardt-sociomantic: Flushing seems not to be necessary here.

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,329 @@
...
+
+                // TODO: The value could be sent in the same message as the
+                // response. For simplicity, we send them separately, at
+                // present.
```
david-eckardt-sociomantic: Why is this simpler?

gavin-norman-sociomantic: It saves having to interpret one message in different ways, on the client side, depending on the contents.

src/dhtproto/node/neo/request/Update.d
david-eckardt-sociomantic: The advantage of `parseBody` is to detect more message protocol errors by ensuring the message has only the expected fields and not additional data. If I understand the protocol for this request correctly, you could use it like this:

```d
auto message = *ed.message_parser.getValue!(MessageType)(this.init_payload);
// ...
with ( MessageType ) switch ( message )
{
    // Normal sequence: read, wait for response, update.
    case GetRecord:
        cstring key;
        ed.message_parser.parseBody(this.init_payload, key);
        response = this.getWaitUpdate(channel, key);
        break;

    // Either an updated record transferred from another node or a
    // new record that did not exist in the DHT before. Just put it.
    case UpdateRecord:
        cstring key;
        hash_t original_hash;
        const!(void)[] new_value;
        // Read original hash and new value here rathern than in `update`
        ed.message_parser.parseBody(this.init_payload, key, original_hash, new_value);
        response = this.update(channel, key, original_hash, new_value);
        break;

    default:
        ed.shutdownWithProtocolError("Unexpected message from client");
}
```

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,329 @@
...
+        // If the record did not exist, end the request.
+        if ( !exists )
+        {
+            ed.send(
+                ( ed.Payload payload )
+                {
+                    payload.addCopy(MessageType.NoRecord);
+                }
+            );
+            ed.flush();
```
david-eckardt-sociomantic: Why not just `return MessageType.NoRecord`?

david-eckardt-sociomantic: It looks like you don't want to end the request, the code comment is wrong then.

src/dhtproto/node/neo/request/Update.d
david-eckardt-sociomantic: I forgot the channel in the above example, but you'll get the point.

src/dhtproto/node/neo/request/Update.d
gavin-norman-sociomantic: Not sure. This looks like a bug, because it shouldn't execute the following section, though it appears that it still will.

src/fakedht/neo/request/Update.d
```diff
@@ -0,0 +1,116 @@
...
+            exist); false if an error occurred
+
+    ***************************************************************************/
+
+    override protected bool get ( cstring channel, hash_t key,
+        void delegate ( Const!(void)[] value ) dg )
+    {
+        auto value_in_channel = global_storage.getCreate(channel).get(key);
+        if ( value_in_channel !is null )
+            dg(value_in_channel);
```
david-eckardt-sociomantic, line 72: Can `value_in_channel` be `null`?

src/dhtproto/node/neo/request/Update.d
david-eckardt-sociomantic: It could be left to the client to decide what to do in this case and respond either with a record value to put in the DHT or a message to cancel the request. This would increase the complexity of the request but be user friendly, avoiding a subsequent `Put` request if the record doesn't exist.

relnotes/update.feature.md
```diff
@@ -0,0 +1,10 @@
+### Neo Update request
+
+`dhtproto.client.DhtClient`, `dhtproto.client.request.Update`
+
+The new request, `Update`, fetches a record value, allows the user to speficy an
```
david-eckardt-sociomantic: Typo `specify`

src/dhtproto/node/neo/request/Update.d
gavin-norman-sociomantic: Yeah, I deliberately chose to leave this as a future enhancement. (An earlier version of Update actually did handle this extra logic, but I removed it to simplify things. I guess this is where this bug / inconsistency comes from.)

gavin-norman-sociomantic:
Updated:
* Fixed bug around `NoRecord`.
* Refactored code to send/receive the message type and record value in a single payload. (Removed TODOs.)
* Fixed typo in release notes.

gavin-norman-sociomantic:
Updated:
* Use `parseBody` in node-side `update` method.

src/dhtproto/client/NotifierTypes.d
```diff
@@ -56,6 +56,50 @@ public struct RequestRecordInfo
...
+    /// Record value.
+    Const!(void)[] value;
+
+    /// Buffer to receive updated value.
+    void[]* updated_value;
+
+    /// Template method to deserialize `value` as a given struct.
+    mixin DeserializeMethod!(value);
+
+    // TODO add serialize method
```
nemanja-boric-sociomantic: Unresolved TODO.

nemanja-boric-sociomantic: Unresolved TODO.

src/dhtproto/client/mixins/NeoSupport.d
```diff
@@ -977,6 +1009,61 @@ template NeoSupport ( )
...
+
+            Params:
+                channel = name of the channel to update a record in
+                key = hash of the record to update
+                user_notifier = notifier delegate
+
+        ***********************************************************************/
+
+        public void update ( cstring channel, hash_t key,
+            Neo.Update.Notifier user_notifier )
```
nemanja-boric-sociomantic, line 1027: Do we want this one to return `bool`?

src/dhtproto/client/request/Update.d
```diff
@@ -0,0 +1,129 @@
...
+        3. If the record exists, its value is passed to the user, who may
+           provide a modified version of the value or decide to leave the value
+           as it is.
+        4. The modified version of the value is then sent back to the DHT. One
+           of the following then happens:
+            4a. The record has been modified by another request, in the
+                meantime. The Update request is rejected and has no effect. (The
+                user may retry the request.)
+            4b. The record has not been modified, in the meantime, and the new
+                value is written to the DHT.
```
nemanja-boric-sociomantic: `The record was deleted or has not been modified`, to be pedant.

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,312 @@
...
+        // Get the currently stored record value and hash it.
+        bool exists;
+        hash_t stored_hash;
+        auto ok = this.get(channel, key,
+            ( Const!(void)[] value )
+            {
+                exists = true;
+                stored_hash = Fnv1a(value);
+            }
+        );
```
nemanja-boric-sociomantic, line 236: I hope there's a guarantee there's no fiber switch between this delegate executing and the next line. Because if there's one, you end up with the race condition within this small window.

src/dhtproto/node/neo/request/Update.d
nemanja-boric-sociomantic: Ping on `parseBody`

nemanja-boric-sociomantic (commented):
Some initial comments.

src/dhtproto/node/neo/request/Update.d
gavin-norman-sociomantic: Hm, it depends on the storage engine's implementation. In the two extant implementations, no fiber switch can occur.

nemanja-boric-sociomantic: Right. Perhaps is enough just to add a comment here.

src/dhtproto/client/mixins/NeoSupport.d
gavin-norman-sociomantic: That would inconsistent with the other blocking requests in this class.

src/dhtproto/client/request/Update.d
gavin-norman-sociomantic: Mm, which sentence is this an addition to?

src/dhtproto/node/neo/request/Update.d
gavin-norman-sociomantic: It can't be used here, as the payload is passed on to `update`, where more fields are read.

src/dhtproto/client/request/Update.d
nemanja-boric-sociomantic: I forgot how the code looked like (I need to check), but AFAIR new value gets written in two cases:

1. the old record value is still valid
2. record ceased to exist in the meantime

src/dhtproto/node/neo/request/Update.d
nemanja-boric-sociomantic: All clear, thanks!

gavin-norman-sociomantic:
Updated:
* Added serialize method to Update notifier.
* Added usage example and test for serialization methods.
* Mentioned deleted records in comment.

src/dhtproto/client/NotifierTypes.d
```diff
@@ -89,3 +134,65 @@ public struct RequestKeyInfo
...
+    ***************************************************************************/
+
+    public void serialize ( T ) ( T src )
+    {
+        static if ( Version.Info!(T).exists )
+        {
+            if ( client_serializer_version_decorator is null )
+                client_serializer_version_decorator = new VersionDecorator;
+
+            client_serializer_version_decorator.store!(T)(src, *dst);
```
joseph-wakeling-sociomantic: I don't think you actually need to bother with allocating a singleton `VersionDecorator` instance: `VersionDecorator.store` will do, as it is a `static` method:
https://github.com/sociomantic-tsunami/ocean/blob/e9b693dc46d6d6b80bd54244b004b0ecd7679c64/src/ocean/util/serialize/contiguous/MultiVersionDecorator.d#L107

The underlying buffer of `VersionDecorator` is only used for deserialization IIRC.

src/dhtproto/client/NotifierTypes.d
```diff
@@ -89,3 +134,65 @@ public struct RequestKeyInfo
...
+    Must be public as it needs to be accessible from the point where the
+    serialize method is mixed in. Not intended to be used externally by users.
+    The object is given an unusually long name so as to reduce the likelihood of
+    name clashes with other symbols.
+
+    TODO: deprecated, replace with implementation in swarm
+
+*******************************************************************************/
+
+public VersionDecorator client_serializer_version_decorator;
```
joseph-wakeling-sociomantic: In line with remarks above about `VersionDecorator.store`, this should not be necessary, so you don't even need to bother to deprecate it ;-)

relnotes/update.feature.md
```diff
@@ -0,0 +1,10 @@
+### Neo Update request
+
+`dhtproto.client.DhtClient`, `dhtproto.client.request.Update`
+
+The new request, `Update`, fetches a record value, allows the user to specify an
+updated value, and replaces the old value in the node with the new value. Note
+that the request will notice if the value being updated has been modified by
+another request, while the Update is in progress. If this happens, the Update
+will be rejected (the client should retry the request).
```
joseph-wakeling-sociomantic, line 9: Feels like the `{De}SerializeMethod` stuff should be discussed here (not necessarily in detail, but the principle of what is available).

src/dhtproto/client/NotifierTypes.d
gavin-norman-sociomantic: Ah that's good to know :+1: .

relnotes/update.feature.md
gavin-norman-sociomantic: Seems a bit specific to me. My feeling is that details of usage should be (and are) covered by the usage examples.

src/dhtproto/client/NotifierTypes.d
gavin-norman-sociomantic: Updated.

relnotes/update.feature.md
joseph-wakeling-sociomantic: TBH I am a little bit mystified as to why swarm/dhtproto are even caring about (de)serialization.  I suppose it makes it simpler to define the delegates without having to provide them a manually-defined context where deserialization functionality is available.  But OTOH it takes us away from the historical assumption that it's for the user to care about what values are, not the swarm protocols.

Is there something specific to the update request that requires this?

gavin-norman-sociomantic: Nope, the protocol doesn't care. It's just a convenience for the user, who would have to implement this anyway. The other DHT neo requests have the same feature.

src/dhtproto/client/NotifierTypes.d
```diff
@@ -89,3 +134,43 @@ public struct RequestKeyInfo
...
+        Params:
+            T = type of struct to serialize
+            src = instance to serialize
+
+    ***************************************************************************/
+
+    public void serialize ( T ) ( T src )
+    {
+        static if ( Version.Info!(T).exists )
+            VersionDecorator.store!(T)(src, *dst);
```
joseph-wakeling-sociomantic, line 169: BTW, with respect to the singleton `VersionDecorator` instance used by `swarm.neo.client.mixins.DeserializeMethod`: given that only `loadCopy` is used, it's not actually necessary for the singleton instance to be exposed directly.  It should suffice to have a `public` function that uses a `private VersionDecorator` instance.

gavin-norman-sociomantic: :+1: that would be a slight improvement.

gavin-norman-sociomantic: (A question for swarm, of course. Not related to this PR.)

src/dhtproto/client/NotifierTypes.d
```diff
@@ -89,3 +134,65 @@ public struct RequestKeyInfo
             this.request_id, this.key);
     }
 }
+
+import ocean.util.serialize.contiguous.MultiVersionDecorator;
```
nemanja-boric-sociomantic: Perhaps move this import to the top of the file.

src/dhtproto/client/NotifierTypes.d
nemanja-boric-sociomantic: This doesn't seem like it's updated?

src/dhtproto/client/NotifierTypes.d
nemanja-boric-sociomantic: ditto - this is not needed if you use `VersionDecorator.store`

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,314 @@
+/*******************************************************************************
+
+    Update request protocol.
+
+    Copyright:
+        Copyright (c) 2018 sociomantic labs GmbH. All rights reserved.
```
nemanja-boric-sociomantic, line 6: RIP

src/dhtproto/node/neo/request/Update.d
nemanja-boric-sociomantic: ^ Not needed.

src/dhtproto/node/neo/request/Update.d
```diff
@@ -0,0 +1,314 @@
...
+
+    ***************************************************************************/
+
+    abstract protected bool responsibleForKey ( hash_t key );
+
+    /***************************************************************************
+
+        Reads a single record from the storage engine. Note that the
+        implementing class must guarantee that no fiber switch can occur during
+        this method.
```
nemanja-boric-sociomantic, line 267: :+1:

src/dhtproto/client/NotifierTypes.d
```diff
@@ -168,31 +168,9 @@ template SerializeMethod ( alias dst )
...
-    Must be public as it needs to be accessible from the point where the
-    serialize method is mixed in. Not intended to be used externally by users.
-    The object is given an unusually long name so as to reduce the likelihood of
-    name clashes with other symbols.
-
-    TODO: deprecated, replace with implementation in swarm
-
-*******************************************************************************/
-
-public VersionDecorator client_serializer_version_decorator;
```
nemanja-boric-sociomantic: Ah, this all looks like it's in the wrong commit, that's why I didn't notice it before.

src/dhtproto/client/request/internal/Update.d
```diff
@@ -0,0 +1,792 @@
...
+                            success = true;
+                            break;
+                        case Abort:
+                            // shared_working.result may either be the default
+                            // value (Error) or may have been set to a specific
+                            // code.
+                            error = true;
+                            break;
+                        case Retry:
+                            break;
```
nemanja-boric-sociomantic: Perhaps add a comment here:

```
// No record on this node, retry on the next one
```

src/dhtproto/client/request/internal/Update.d
```diff
@@ -0,0 +1,792 @@
...
+                            this.context.shared_working.result =
+                                Update.SharedWorking.Result.Succeeded;
+                            success = true;
+                            break;
+                        case Abort:
+                            // shared_working.result may either be the default
+                            // value (Error) or may have been set to a specific
+                            // code.
+                            error = true;
+                            break;
```
nemanja-boric-sociomantic, line 294: It's weird that this `Abort` actually doesn't abort anything - perhaps we should rename it to something else?

src/dhtproto/client/request/internal/Update.d
```diff
@@ -0,0 +1,792 @@
...
+    }
+
+    /***************************************************************************
+
+        Starts a new request-on-conn to write the record value to another node.
+        Asks the node the record was read from to remove it; handles the
+        response code.
+
+        Params:
+            conn = connection to communicate over
```
nemanja-boric-sociomantic: `connection to communicate with the node that originally had the record`?

src/dhtproto/client/request/internal/Update.d
```diff
@@ -0,0 +1,792 @@
...
+            case Ok:
+                return true;
+
+            case Error:
+                // The node returned an error code. Notify the user.
+                Update.Notification n;
+                n.node_error = RequestNodeInfo(this.context.request_id,
+                    conn.remote_address);
+                Update.notify(this.context.user_params, n);
+                return false;
```
nemanja-boric-sociomantic, line 614: Should we maybe add a `FIXME` here - what happens if we write the new record to the newest node, but we fail to remove it from the old node (perhaps because of the connection issue)? Will we get in any kind of trouble for leaving the record on both nodes? Should we remove it from the new node as well?

src/dhtproto/client/UsageExamples.d
```diff
@@ -878,6 +878,191 @@ unittest
...
+                    auto deserialized = info.received.deserialize(record);
+                    this.log.trace("Deserialized: {} / {} / {}",
+                        deserialized.name, deserialized.id,
+                        deserialized.daily_totals);
+
+                    // Update the record.
+                    deserialized.daily_totals[0]++;
+
+                    // Serialize the updated record.
+                    info.received.serialize(*deserialized);
```
nemanja-boric-sociomantic, line 982: Nit pick: It feels a bit odd that you do not reference updated_value here.

nemanja-boric-sociomantic (commented):
In general LGTM, although I have some questions/suggestions.

src/dhtproto/client/UsageExamples.d
gavin-norman-sociomantic: You mean `info.received.updated_value`? There's no need to, if you use the `serialize` method.

nemanja-boric-sociomantic: Yes, but it did caught me off guard - it _feels_ like you're serializing into the received record again, and not it's updated value. I'm not sure what's best here - I think some kind of naming where this `received` would be just `record` _might_ be better.

src/dhtproto/client/request/internal/Update.d
gavin-norman-sociomantic: It won't be a problem because when the redistribution process moves the record from the old -> new node, it will be discarded as the record already exists on the new node.

nemanja-boric-sociomantic: That's good, thanks for the clarification!

src/dhtproto/client/UsageExamples.d
gavin-norman-sociomantic: Hmm, yeah, `received` here represents the notification "we have received a record" (i.e. the action), rather than the thing that has been received. I don't think it's worth changing the status quo for the sake of this one odd case.

nemanja-boric-sociomantic: Hm right, I forgot about that. 👍

src/dhtproto/node/neo/request/Update.d
gavin-norman-sociomantic: I'll leave this for a batch transformation.

gavin-norman-sociomantic:
Updated.

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
I've restarted the build, some issues with the apt repos are in play.

PR #116, opened 2018-02-20
New legacy channel mirror implementation

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Needs:
1. Release notes.
2. Testing in a real app.

gavin-norman-sociomantic:
Added an extra commit with a unittest to check that the new `ExtensibleMirror` template compiles.

src/dhtproto/client/legacy/internal/helper/ExtensibleChannelMirror.d
```diff
@@ -59,13 +59,15 @@ import dhtproto.client.DhtClient;
...
+            MirrorBase!(Dht)
         Plugins = variadic list of record-handling plugins to be called (*in the
             order specified*) when records are received by the mirror
 
 *******************************************************************************/
 
-public class ExtensibleChannelMirror ( Dht : DhtClient, Plugins ... )
-    : ChannelMirror!(Dht)
+public class ExtensibleMirror
+    ( Dht : DhtClient, MirrorImpl : MirrorBase!(Dht), Plugins ... ) : MirrorImpl
```
joseph-wakeling-sociomantic, line 70: I wonder if you could reasonably avoid the boilerplate-y requirement to specify both `Dht` and `MirrorImpl` here.  The `MirrorImpl` comes with the DHT client type defined, so there must be some other means to validate that it is indeed a subclass of `MirrorBase` ... ?

gavin-norman-sociomantic: Yeah, good point. I'll have a look. It might be possible.

joseph-wakeling-sociomantic: Minor gripe, of course.  The new setup seems to be working very nicely in general.

src/dhtproto/client/legacy/internal/helper/Mirror.d
```diff
@@ -0,0 +1,527 @@
...
+            {
+                this.log.info("Starting.");
+                this.outer.dht.assign(rq);
+                this.state = State.Assigned;
+            }
+            else
+            {
+                this.log.info("Starting in {}ms.", start_in_ms);
+                this.outer.dht.schedule(rq, start_in_ms);
+                this.state = State.Scheduled;
```
joseph-wakeling-sociomantic, line 115: Minor note: this means that one never sees any log output for when a `GetAll` request actually starts, bar the initial one.  One just sees the `Starting in {}ms.` message.

joseph-wakeling-sociomantic:
I've run local tests with this in 3 apps:

* one that uses `ChannelMirror`, where it was easy to do a quick `s/ChannelMirror/Mirror/` and everything seems to work fine (following the fix to checks in the `start` method)

* two that use `ExtensibleChannelMirror`, where I tested first without changing any app code (to make sure that the `ExtensibleChannelMirror` template in the new implementation works), and second switching the mirror instances out with `ExtensibleMirror!(Dht, Mirror!(Dht), ...)`.  Both cases seem to work fine.

  * in one of these apps, I even tried out a scenario where some of the mirrors were using `ExtensibleMirror` and some `ExtensibleChannelMirror`, and this also worked fine.

src/dhtproto/client/legacy/internal/helper/ExtensibleChannelMirror.d
gavin-norman-sociomantic: I can't see a simple way of doing this. I'll leave it as it is; I don't think it's worth spending time reworking stuff.

joseph-wakeling-sociomantic: No problem.  It's a small price to pay for a really nice improvement in how legacy mirrors work.

joseph-wakeling-sociomantic:
I've also now tried this out in a partial-handshake scenario, and I think there may be an issue with the periodic `GetAll` requests launched by the `mirror()` method.

The broad scenario here is that a client is started up with only a subset of DHT nodes available.  Then we see:

* the app waits until at least one DHT node is connected, and times out waiting for all of them

* an initial `Mirror.fill()` call is issued, which launches a `GetAll` request that succeeds of course only on the nodes currently connected

* the app times out waiting for the fill to succeed on all nodes, and calls `mirror()`

* `Listen` and periodic `GetAll` requests are scheduled successfully for the connected nodes

* the missing DHT nodes are launched

* no periodic `GetAll` requests seem to occur for these nodes, although it looks like the `GetAll` requests launched by `fill()` finally complete successfully.

At a guess I would assume the fact that the `fill()`-launched `GetAll` has not succeeded prevents the `mirror()`-launched periodic `GetAll` from being scheduled at all for these nodes.  Let's dig deeper tomorrow ;-)

joseph-wakeling-sociomantic:
Updated version (where the `mirror` request converts any existing one-shot `GetAll`s to repeating) fixes the issue AFAICT.

src/dhtproto/client/legacy/internal/helper/mirror/model/MirrorBase.d
```diff
@@ -0,0 +1,257 @@
...
+            RequestScheduler plugin)
+
+*******************************************************************************/
+
+abstract public class MirrorBase ( Dht : DhtClient )
+{
+    /// This helper only works in conjunction with the request scheduler plugin,
+    /// so we assert that that exists in the client.
+    static assert(Dht.HasPlugin!(DhtClient.RequestScheduler),
+        "ChannelMirror helper requires RequestScheduler plugin in client");
```
joseph-wakeling-sociomantic: Should this say `ChannelMirror` ... ?

src/dhtproto/client/legacy/internal/helper/Mirror.d
```diff
@@ -0,0 +1,533 @@
...
+        GetAll finished notification.
+
+        The base class does nothing, but this method may be overridden by
+        derived classes to implement special fill behaviour.
+
+    ***************************************************************************/
+
+    protected void fillFinished ( )
+    {
+    }
```
joseph-wakeling-sociomantic, line 478: Why is `fillFinished` defined only in `ChannelMirror` and `Mirror`, rather than being part of the base?

joseph-wakeling-sociomantic (commented):
Going by experience, this LGTM.  Just couple of small comments that you might want to consider ...

src/dhtproto/client/legacy/internal/helper/mirror/model/MirrorBase.d
gavin-norman-sociomantic: Ah yes, copy pasta.

src/dhtproto/client/legacy/internal/helper/Mirror.d
gavin-norman-sociomantic: Because the base class has no way of knowing when this method should be called, and no way of enforcing that derived classes must call it when appropriate.

joseph-wakeling-sociomantic: Fair enough.  I guess it's reasonable that anyone using an actual `MirrorBase` implementation knows which particular subclass they are using, and can rely on its extra features (i.e. one would not use `MirrorBase` directly in an app).

gavin-norman-sociomantic:
Updated:
* Fixed copy pasta.
* Added release notes.

Unblocked.

joseph-wakeling-sociomantic:
Should there not be some release notes related to the introduction of `MirrorBase` and the reworking of `ChannelMirror`?  Or is the goal here to keep that quiet and remove `MirrorBase` and `ChannelMirror` in the next major release?

gavin-norman-sociomantic:
`MirrorBase` is an implementation detail. It's not relevant to users of the DHT client. And yes, the traditional channel mirror will probably be deprecated and removed in the future.

joseph-wakeling-sociomantic (approved):
LGTM, then ;-)

gavin-norman-sociomantic:
Rebased.

gavin-norman-sociomantic:
Added an extra commit that adds the `error` method that exists in the old channel mirror but was not present in the new one. For testing. (To be squashed, if it works correctly.)

gavin-norman-sociomantic:
Blocked until the new commit is squashed or removed.

joseph-wakeling-sociomantic:
Do you need any more application testing?

gavin-norman-sociomantic:
>Do you need any more application testing?

Did you try to adapt an app of yours to use this PR? (Just to check that it compiles. No need for live testing.)

joseph-wakeling-sociomantic:
Well, yes -- please see above comments from ages back ;-)  I'm asking whether I should do so again with the latest code fixes.

gavin-norman-sociomantic:
Ok. Maybe a quick recompile, just for a sanity check.

joseph-wakeling-sociomantic:
OK, I'll try to get it done some time today.

joseph-wakeling-sociomantic:
Checked with one app -- seems OK as ever (though I didn't get as far as checking partial handshake support).

gavin-norman-sociomantic:
Thanks!

joseph-wakeling-sociomantic:
FYI: you can use `fixup! Title of commit you want to modify` as the title for commits like this, and then `git rebase --autosquash` will automatically squash things correctly.

`fixup!` => just squash the code together

`squash!` => squash the code together and combine the commit messages (giving you a chance to edit)

gavin-norman-sociomantic:
I know, thanks. (The commit was added in a hurry for someone to test ;)

gavin-norman-sociomantic:
Updated: squashed `error` commit.

gavin-norman-sociomantic:
Unblocked.

src/dhtproto/client/legacy/internal/helper/mirror/model/MirrorBase.d
```diff
@@ -0,0 +1,257 @@
...
+    /// This helper only works in conjunction with the request scheduler plugin,
+    /// so we assert that that exists in the client.
+    static assert(Dht.HasPlugin!(DhtClient.RequestScheduler),
+        "MirrorBase requires a DHT client with the RequestScheduler plugin");
+
+    /// DHT client used to access DHT.
+    protected Dht dht;
+
+    /// Name of DHT channel to mirror.
+    protected istring channel_;
```
nemanja-boric-sociomantic, line 48: Is there any reason for this to be protected and not private?

gavin-norman-sociomantic: To maintain the API. User code may exist that accesses [`ChannelMirror.channel_`](https://github.com/gavin-norman-sociomantic/dhtproto-tsunami/blob/330ebffb8d2101b7884ede16db9d66c3a48f5dbe/src/dhtproto/client/legacy/internal/helper/ChannelMirror.d#L284) in a derived class. `Mirror` is supposed to be a drop-in replacement for `ChannelMirror`, so the field must have the same protection attribute.

src/dhtproto/client/legacy/internal/helper/mirror/model/MirrorBase.d
```diff
@@ -0,0 +1,257 @@
...
+        Params:
+            setup = struct instance describing the requests to start (defaults
+                to a repeating GetAll request and a continuous Listen)
+
+    ***************************************************************************/
+
+    public void start ( Setup setup = Setup.init )
+    {
+        assert(!(setup.get_all_mode == setup.get_all_mode.None &&
+            setup.listen_mode == setup.listen_mode.None),
```
nemanja-boric-sociomantic: Apply De-Morgan's law here and get: `setup.get_all_mode != setup.get_all_mode.None || setup_listen_mode != setup.listen_mode.None`. IMHO it feels more natural.

src/dhtproto/client/legacy/internal/helper/mirror/model/MirrorBase.d
nemanja-boric-sociomantic: OK, that's a valid reason.

src/dhtproto/client/legacy/internal/helper/mirror/model/MirrorBase.d
gavin-norman-sociomantic: Updated.

relnotes/mirror.feature.md
```diff
@@ -20,3 +20,21 @@ Notes:
...
+```D
+alias ExtensibleChannelMirror!(SchedulingDhtClient,
+    RawRecordDeserializer!(Record), DeserializedRecordCache!(Record))
+    ExampleMirror;
+```
+should be changed to
+```D
+alias ExtensibleMirror!(SchedulingDhtClient, Mirror!(SchedulingDhtClient),
+    RawRecordDeserializer!(Record), DeserializedRecordCache!(Record))
+    ExampleMirror;
```
nemanja-boric-sociomantic, line 38: Is this change worth deprecation of `ExtensibleChannelMirror`?

src/dhtproto/client/legacy/internal/helper/Mirror.d
```diff
@@ -0,0 +1,553 @@
...
+                        info.message(this.outer.dht.msg_buf),
+                        this.outer.retry_time_ms);
+                    this.start(this.outer.retry_time_ms);
+                }
+            }
+        }
+    }
+
+    /// User-provided request notifier.
+    private Dht.RequestNotification.Callback user_notifier;
```
nemanja-boric-sociomantic, line 319: I don't think it matters that user will not be able to tell if the Listen and GetAll request is firing the notifier, right?

nemanja-boric-sociomantic (approved):

relnotes/mirror.feature.md
gavin-norman-sociomantic: Well, usage is intended to be entirely replaced, gradually. I get the feeling that a deprecation may be a _bit_ too much, at this stage, though, as that implies to many people "stop using this code ASAP", which is not quite the case here. We need to be careful, as the new code may of course contain new bugs.

src/dhtproto/client/legacy/internal/helper/Mirror.d
gavin-norman-sociomantic: This is the same in the old channel mirror.

gavin-norman-sociomantic: (Apparently it's never been a problem.)

relnotes/mirror.feature.md
nemanja-boric-sociomantic: Yes, I share the same feeling.

src/dhtproto/client/legacy/internal/helper/Mirror.d
nemanja-boric-sociomantic: Thanks!

relnotes/mirror.feature.md
gavin-norman-sociomantic: Tempting as it is to be gung ho about it and kill off the old one ;)

PR #120, opened 2018-03-05
Enable errors for deprecations and fix those

mihails-strasuns-sociomantic:
Bunch of 2.071 deprecations were not noticed previously because dhtproto
was not compiled with -de flag.

Should be backwards compatible fix AFAICS but please recheck.

### DIFF
--- a/src/dhtproto/client/DhtClient.d
+++ b/src/dhtproto/client/DhtClient.d
@@ -748,7 +748,7 @@ public class DhtClient : IClient
 
     ***************************************************************************/
 
-    private struct Put
+    public struct Put
     {
         mixin RequestBase;
         mixin IODelegate;       // io(T) method
@@ -788,7 +788,7 @@ public class DhtClient : IClient
 
     ***************************************************************************/
 
-    private struct Get
+    public struct Get
     {
         mixin RequestBase;
         mixin IODelegate;       // io(T) method
@@ -826,7 +826,7 @@ public class DhtClient : IClient
 
     ***************************************************************************/
 
-    private struct Exists
+    public struct Exists
     {
         mixin RequestBase;
         mixin IODelegate;       // io(T) method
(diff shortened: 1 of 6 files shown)


src/dhtproto/client/DhtClient.d
```diff
@@ -748,7 +748,7 @@ public class DhtClient : IClient
 
     ***************************************************************************/
 
-    private struct Put
+    public struct Put
```
gavin-norman-sociomantic, line 751: It's a shame to have to make this change. It shouldn't be possible to create instances of this struct except via the associated method. I suppose there's no way to enforce this now in a D1/D2 compatible way.

mihails-strasuns-sociomantic:
FYI @mathias-lang-sociomantic

gavin-norman-sociomantic (approved):

mathias-lang-sociomantic:
This branch is not tested with v2.071.0

mihails-strasuns-sociomantic:
@mathias-lang-sociomantic I wrote it first for v14.x.x branch, tested there, and rebased to v13.2.x after. AFAIR we can't enable 2.071 for this branch because of submodules but code in dhtproto itself is compatible with this PR.

src/dhtproto/client/DhtClient.d
mihails-strasuns-sociomantic: Yes, you can eventually mark struct constructor as `@disable this()` in D2. There are still quite some ways to create wrong instance (i.e. `T.init`) but that will cover "accidental" ones.

Here I had to mark it as public because you actually do use this type from other module.

mathias-lang-sociomantic (approved):

src/dhtproto/client/DhtClient.d
gavin-norman-sociomantic: Ahh something internal in dhtproto, right. (Do you recall which module that was btw?)

mihails-strasuns-sociomantic: No, but it was something defining aliases with those types as template arguments

PR #123, opened 2018-03-14
Sanity check: Merge v14.x.x -> neo

gavin-norman-sociomantic:
(Based on #122.)


gavin-norman-sociomantic:
Recreated after merging #122.

gavin-norman-sociomantic:
Re-pushed with import fixes for dmd 2.071.

gavin-norman-sociomantic:
The neo branch needs some serious dmd 2.071 fix-ups. Blocked until #124 is merged.

gavin-norman-sociomantic:
Updated and unblocked.

gavin-norman-sociomantic:
One build failed due to trying to start the fake node on a port that was already in use. Looking through the code, it appears that:
* The main fakedht test uses a fixed port (10,000).
* All other tests use a randomly assigned port.
* The neo port is only [hardcoded (legacy port + 100) if the legacy port is a fixed number](https://github.com/sociomantic-tsunami/dhtproto/blob/neo/src/fakedht/DhtNode.d#L100-L101). This means that if the legacy port is assigned to an ephemeral port, the neo port will be too.

Only one of the four tests is using a fixed port, so running tests in parallel _should_ be fine. It's not clear to me how two tests could end up using the same port :thinking:

gavin-norman-sociomantic:
I've restarted the tests a few times and they've passed. We've never seen this port conflict before, afair, so I think this is a pretty "ephemeral" problem. I'll go ahead and merge. We can investigate possible port conflict flakiness separately.

nemanja-boric-sociomantic:
That sounds good to me. If this starts to happen again, we can look into it.

Issue #126, opened 2018-03-20
Partial handshake support: Add Task-blocking `RetryHandshake`


gavin-norman-sociomantic:
Either a new implementation or a derived class.

Some applications already have code for this, so it's probably just a case of surveying what we already have, checking it's generic enough for a library, and bringing it in.

gavin-norman-sociomantic:
Some apps are already adapted to use the existing `RetryHandshake` in a task-blocking manner. Code like the following indicates that this is simple:
```D
    override public void waitForFullHandshake ( )
    {
        auto task = Task.getThis();
        assert(task !is null);

        bool done;
        new RetryHandshake(theScheduler.epoll, this.dht, RETRY_WAIT, {
            done = true;
            if ( task.suspended ) task.resume;
        });

        if ( !done ) task.suspend();
    }
```

Given the simple usage of `RetryHandshake` with tasks and the fact that the exact behaviour required for partial handshakes varies greatly by application, I'm not sure it's worth adding anything more to dhtproto.

PR #129, opened 2018-03-21
Add new helper class to wrap RetryHandshake with task support

joseph-wakeling-sociomantic:
The new DhtHandshake class is designed to make it easy for applications to implement support for a partial DHT handshake, e.g.:

```D
auto handshake = new DhtHandshake(theScheduler.epoll, dht_client, 10);

// block until at least one node is connected
theScheduler.await(handshake.oneNodeConnected());

// wait until either all nodes have connected, or 60 seconds
// have passed, whichever comes sooner
theScheduler.awaitOrTimeout(handshake.allNodesConnected(), 60_000_000);
```

The underlying `RetryHandshake` instance should ensure that even if the timeout is hit in the above example, the application will keep trying to contact any missing nodes until it succeeds.

Simple logging has been built into the class in order to make it easy to monitor handshake progress.

Fixes https://github.com/sociomantic-tsunami/dhtproto/issues/126.

### DIFF
--- a/src/dhtproto/client/legacy/internal/helper/Handshake.d
+++ b/src/dhtproto/client/legacy/internal/helper/Handshake.d
@@ -1,0 +1,241 @@
+/*******************************************************************************
+
+    Class to perform a handshake with all DHT nodes (retrying when it fails),
+    providing task helpers that block until one or all nodes are connected.
+
+    Copyright: Copyright (c) 2018 sociomantic labs GmbH.  All rights reserved
+
+    License: Boost Software License Version 1.0.  See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module dhtproto.client.legacy.internal.helper.Handshake;
+
+/// ditto
+public class DhtHandshake
+{
+    import dhtproto.client.DhtClient;
+    import dhtproto.client.legacy.internal.helper.RetryHandshake;
+
+    import ocean.task.Scheduler;
+    import ocean.task.Task;
+    import ocean.task.util.Event;
+    import ocean.util.log.Logger;
+
+
+    /***************************************************************************
+
+        Task that suspends until all nodes are connected
+
+    ***************************************************************************/
+
+    private class AllDhtNodesConnected : Task
+    {
+        /// Used to manage suspend/resume of this task
+        private TaskEvent all_connected_event;
+
+        /***********************************************************************
+
+            Delegate to be called by the handshake once all DHT nodes have
+            been reached
+
+        ***********************************************************************/
+
+        private void allNodesConnected ()
+        {
+            this.outer.log.info("All DHT nodes connected!");
+            this.all_connected_event.trigger();
+        }
+
+        /***********************************************************************
+
+            Runs the task
+
+        ***********************************************************************/
+
+        protected override void run ()
+        {
+            this.all_connected_event.wait();
+        }
+    }
+
+
+    /***************************************************************************
+
+        Task that suspends until at least one node is connected
+
+    ***************************************************************************/
+
+    private class OneDhtNodeConnected : Task
+    {
+        import swarm.Const : NodeItem;
+
+        /// Used to manage suspend/resume of this task
+        private TaskEvent node_connected_event;
+
+        /***********************************************************************
+
+            Delegate to be called by the handshake each time an individual
+            node connects
+
+            Params:
+                item = data on the address and port of the connected node
+
+        ***********************************************************************/
+
+        private void oneNodeConnected (NodeItem item)
+        {
+            this.outer.log.trace("Connected to DHT node {}:{}",
+                                 item.Address, item.Port);
+            this.node_connected_event.trigger();
+        }
+
+        /***********************************************************************
+
+            Runs the task
+
+        ***********************************************************************/
+
+        protected override void run ()
+        {
+            this.node_connected_event.wait();
+        }
+    }
+
+
+    /// Logger for the handshake
+    private Logger log;
+
+
+    /// Instance of task that suspends until all nodes are connected
+    private AllDhtNodesConnected all_nodes_connected;
+
+
+    /// Instance of task that suspends until at least one node is connected
+    private OneDhtNodeConnected one_node_connected;
+
+
+    /// Instance of class that launches the actual handshake, retrying when
+    /// it fails until all nodes are connected
+    private RetryHandshake retry_handshake;
+
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            dht = DHT client instance via which to perform the handshake
+            retry_delay_seconds = time to wait (in seconds) before retrying
+                failed handshake attempts
+
+    ***************************************************************************/
+
+    public this (DhtClient dht, size_t retry_delay_seconds)
+    in
+    {
+        assert(dht !is null);
+    }
+    body
+    {
+        this.log = Log.lookup("dhtproto.client.legacy.internal.helper.Handshake");
+
+        this.all_nodes_connected = new AllDhtNodesConnected;
+        this.one_node_connected = new OneDhtNodeConnected;
+
+        this.retry_handshake =
+            new RetryHandshake(theScheduler.epoll, dht, retry_delay_seconds,
+                    &this.all_nodes_connected.allNodesConnected,
+                    &this.one_node_connected.oneNodeConnected);
+    }
+
+
+    /***************************************************************************
+
+        Returns:
+            instance of task that suspends until all nodes have been connected
+
+    ***************************************************************************/
+
+    public AllDhtNodesConnected allNodesConnected ()
+    {
+        return this.all_nodes_connected;
+    }
+
+
+    /***************************************************************************
+
+        Returns:
+            instance of task that suspends until at least one node is connected
+
+    ***************************************************************************/
+
+    public OneDhtNodeConnected oneNodeConnected ()
+    {
+        return this.one_node_connected;
+    }
+}
+
+///
+unittest
+{
+    // this example assumes we are already running in
+    // a task context
+    class HandshakeExample
+    {
+        import dhtproto.client.DhtClient;
+
+        import ocean.io.Stdout;
+        import ocean.task.Scheduler;
+
+        // DHT client to connect with
+        private DhtClient dht;
+
+        // The `DhtHandshake` instance may need to persist beyond the
+        // end of the `run` method (if e.g. we complete only a partial
+        // handshake by the time it exits)
+        private DhtHandshake handshake;
+
+
+        // Set up a DHT client and initialize the handshake, blocking
+        // until either the handshake completes, or until at least a
+        // partial handshake is complete and more than 60 seconds have
+        // passed
+        void run ()
+        {
+            // we assume `theScheduler` is already initialized
+            // as this is used internally by `DhtHandshake`
+            this.dht = new DhtClient(theScheduler.epoll);
+
+            // start the handshake
+            auto retry_delay_seconds = 3;
+            this.handshake = new DhtHandshake(this.dht, retry_delay_seconds);
+
+            // block on at least one node connecting
+            theScheduler.await(this.handshake.oneNodeConnected());
+            Stdout.formatln("At least one node is now connected!");
+
+            // if `theScheduler.awaitOrTimeout` is available, it can
+            // be used to implement support for a partial handshake:
+            //
+            // ```
+            // auto timeout_microsec = 60_000_000;
+            //
+            // auto handshake_timeout =   /* true if timeout is reached */
+            //     theScheduler.awaitOrTimeout(
+            //         this.handshake.allNodesConnected(),
+            //         timeout_microsec);
+            //
+            // /* react to `handshake_timeout` as you wish, but note
+            //    that the `DhtHandshake` instance will keep working
+            //    in the background to complete the handshake */
+            // ```
+            //
+            // alternatively, we can block on all DHT nodes being
+            // connected using `await`:
+            theScheduler.await(this.handshake.allNodesConnected());
+
+            Stdout.formatln("DHT handshake complete!");
+        }
+    }
+}
--- a/relnotes/handshake.feature.md
+++ b/relnotes/handshake.feature.md
@@ -1,0 +1,42 @@
+### New handshake helper with built-in task support
+
+`dhtproto.client.legacy.internal.helper.Handshake`
+
+The new `DhtHandshake` class wraps the existing `RetryHandshake` with
+easy support for a task-based workflow.  This should make it easy for
+applications to support a partial handshake, e.g.:
+
+```D
+auto retry_delay_seconds = 3;
+auto handshake = new DhtHandshake(dht_client, retry_delay_seconds);
+
+// block on at least one node connecting
+theScheduler.await(handshake.oneNodeConnected());
+Stdout.formatln("At least one node is now connected!");
+
+// wait until either all nodes have connected, or 60 seconds
+// have passed, whichever comes sooner (N.B. `awaitOrTimeout`
+// is only available for more recent ocean releases)
+auto timeout_microsec = 60_000_000;
+
+auto handshake_timeout =   // true if timeout is reached
+    theScheduler.awaitOrTimeout(
+        handshake.allNodesConnected(),
+        timeout_microsec);
+
+if (handshake_timeout)
+{
+    Stdout.formatln(
+        "DHT handshake did not succeed within {} seconds!",
+        timeout_microsec / 1_000_000);
+}
+else
+{
+    Stdout.formatln(
+        "DHT handshake reached all nodes before timeout!");
+}
+
+// if we timed out, the `DhtHandshake` instance will still
+// keep working in the background to connect to the missing
+// DHT nodes, so all nodes should be reached eventually
+```


joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic does this fit what you had in mind? :-)

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+
+        /***********************************************************************
+
+            Runs the task
+
+        ***********************************************************************/
+
+        protected override void run ()
+        {
+            if (!this.all_dht_nodes_connected) this.suspend();
```
mihails-strasuns-sociomantic: Can also just use  `TaskEvent` here

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+            node connects
+
+        ***********************************************************************/
+
+        private void dg (NodeItem item)
+        {
+            if (!this.at_least_one_node_connected)
+            {
+                this.at_least_one_node_connected = true;
+                if (this.suspended()) this.resume();
```
mihails-strasuns-sociomantic: ditto `taskEvent`

mihails-strasuns-sociomantic (approved):
Yup, pretty much how I have expected it too look!

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+        private bool all_dht_nodes_connected = false;
+
+        /***********************************************************************
+
+            Delegate to be called by the handshake once all DHT nodes have
+            been reached
+
+        ***********************************************************************/
+
+        private void dg ()
```
gavin-norman-sociomantic: Would be a bit clearer to call this `allNodesConnected`.

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+        private bool at_least_one_node_connected = false;
+
+        /***********************************************************************
+
+            Delegate to be called by the handshake each time an individual
+            node connects
+
+        ***********************************************************************/
+
+        private void dg (NodeItem item)
```
gavin-norman-sociomantic: Would be a bit clearer to call this `oneNodeConnected`.

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+
+    /***************************************************************************
+
+        Logger for the handshake
+
+    ***************************************************************************/
+
+    private Logger log;
```
gavin-norman-sociomantic: Optional style point: we tend to prefer `///` comments for everything expect functions, these days.

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+
+    private RetryHandshake retry_handshake;
+
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            epoll = epoll select dispatcher via which to schedule handshake
```
gavin-norman-sociomantic: If this whole thing is based on tasks, is there any reason to accept an epoll selector here? You could just use `theScheduler.epoll`, internally.

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,211 @@
...
+        Returns:
+            instance of task that suspends until at least one node is connected
+
+    ***************************************************************************/
+
+    public OneDhtNodeConnected oneNodeConnected ()
+    {
+        return this.one_node_connected;
+    }
+}
```
gavin-norman-sociomantic, line 177: Please add a usage example (like what you included in the commit message).

gavin-norman-sociomantic:
Generally, this seems like a nice, convenient, little helper :+1:

src/dhtproto/client/legacy/internal/helper/Handshake.d
joseph-wakeling-sociomantic: So basically `run` just contains `TaskEvent.wait();`, and the soon-to-be-renamed `dg` just contains any wanted logging plus `TaskEvent.trigger();` ... ?

src/dhtproto/client/legacy/internal/helper/Handshake.d
joseph-wakeling-sociomantic: I didn't want to hardcode an unnecessary dependency on `theScheduler`, on the grounds that the user might have a different setup (e.g. custom `Scheduler` instance instead of `theScheduler`).  But that might be over-engineering.

src/dhtproto/client/legacy/internal/helper/Handshake.d
joseph-wakeling-sociomantic: Yup, will do so.  Just wanted to get the design itself tied down/approved first.

How feasible is it to really _test_ something like this?  I don't have much experience of swarm/proto unittesting and integration tests.

src/dhtproto/client/legacy/internal/helper/Handshake.d
mihails-strasuns-sociomantic: Yup

mihails-strasuns-sociomantic: In that sense it just occured  to me that whole handshake helper could be a task event, and not actual task - but it doesn't matter much and this way probably is more generic.

src/dhtproto/client/legacy/internal/helper/Handshake.d
mihails-strasuns-sociomantic: > e.g. custom Scheduler instance instead of theScheduler

This is not supported by ocean

gavin-norman-sociomantic:
The commit message can say "fixes #126", btw.

Oh: release notes! :police_car:

src/dhtproto/client/legacy/internal/helper/Handshake.d
joseph-wakeling-sociomantic: By whole handshake helper you mean the `DhtHandshake` class?  It's not a task itself, it just exposes a couple.

If I understand you right, you mean we could just use a couple of `TaskEvent` instances under the hood and expose methods that would cause the calling task context to invoke their `wait()` ... ?

FWIW I agree that it's nicer/more flexible to expose task instances.  But I don't mind drafting an alternative design (in a separate PR) for comparison.

joseph-wakeling-sociomantic: Probably it could be something like:

```D
class DhtHandshake
{
    private RetryHandshake retry_handshake;

    private TaskEvent one_node_connected;

    private TaskEvent all_nodes_connected;

    public bool start (DhtClient dht, uint timeout, size_t wait_time)
    {
        verify(this.retry_handshake is null);

        void oneNodeConnectedDg (NodeItem item)
        {
            this.one_node_connected.trigger();
        }

        bool timed_out = true;

        void allNodesConnectedDg ()
        {
            timed_out = false;
            this.all_nodes_connected.trigger();
        }

        this.retry_handshake =
            new RetryHandshake(theScheduler.epoll, dht,
                    &allNodesConnectedDg, &this.oneNodeConnectedDg);

        this.one_node_connected.wait();

        if (timeout > 0)
        {
            // launch some timer event that will also call
            // this.all_nodes_connected.trigger()
        }

        this.all_nodes_connected.wait();

        return timed_out;
    }
}
```

... no ... ?

mihails-strasuns-sociomantic: Almost what I had in mind - but you don't need to call `this.one_node_connected.wait();` there, it will be called from host ask (same for `this.all_nodes_connected.wait();`). Benefit with this approach is that you don't need to use 2 worker fibers for one/all cases, it will only suspend one host task - also allocates slightly less memory. Drawback is that it seems slightly more convoluted.

joseph-wakeling-sociomantic: Hmm, are you expecting that the `TaskEvent` fields will be public?  I don't see any other way to viably do what you have in mind given that they are struct instances.  And my inclination was to hide the `trigger` method at least.

FWIW I do think that on balance the existing version using tasks, while a little less efficient (in LoC as well as memory) offers a slightly more intuitive UI.  But do you think it would be worth using `TaskEvent` under the hood of the `{One,All}Node{s}Connected` tasks, in any case?

mihails-strasuns-sociomantic: > Hmm, are you expecting that the TaskEvent fields will be public? I 

You only need to expose public methods like `wait_one_node()` which calls event internally, actual event struct instance is of no other use anyway.

mihails-strasuns-sociomantic: > But do you think it would be worth using TaskEvent under the hood 

Definitely, it easier than to manually look where you have to add `if (suspended)` checks

joseph-wakeling-sociomantic: > You only need to expose public methods like wait_one_node() which calls event internally, actual event struct instance is of no other use anyway.

That sounds like it'll make the timeout less intuitive, though.  There's something really nice about it being in the hands of the user to say: `await(handshake.oneNode())`, `awaitOrTimeout(handshake.allNodes(), ...)`.  So, I think I'll keep the existing public API, but use `TaskEvent` under the hood of the tasks exposed.  SGTY?

mihails-strasuns-sociomantic: Yeah, as I have mentioned before, I am simply calling out loud available options. Current API looks reasonable.

joseph-wakeling-sociomantic:
Updated to use `TaskEvent` and to rename the `dg` delegates.  I'll follow up with release notes and a unittest usage example.

joseph-wakeling-sociomantic:
... oh, and to remove the `epoll` constructor parameter and use `theScheduler.epoll` instead.

joseph-wakeling-sociomantic:
Release notes, usage examples, and issue cross-refs to follow ;-)

src/dhtproto/client/legacy/internal/helper/Handshake.d
mihails-strasuns-sociomantic: Still missing usage example

mihails-strasuns-sociomantic (commented):
Also red ;)

joseph-wakeling-sociomantic:
Turns out we have a small problem with a usage example: `dhtproto`'s ocean dependency predates the introduction of `Scheduler.awaitOrTimeout` ;-)

joseph-wakeling-sociomantic:
Usage example pushed.  I can see two ways to deal with this:

* bork the usage example to only use `await` (and fix it in dhtproto 14.x.x)

  * maybe noting the possibility in comments

* use ocean versioning to select whether or not to use `awaitOrTimeout`

Any preferences?

mihails-strasuns-sociomantic:
I think first option is better(noting alternative in comments) - version blocks in usage example feel real confusing.

joseph-wakeling-sociomantic:
OK. I will just use `await` and note in a comment that `awaitOrHandshake` can be used (where available) to implement a partial handshake.  And then once this is merged into 14.x.x I'll make a follow-up PR to improve things there.

joseph-wakeling-sociomantic:
Updated with to more to-squash patches:

* fixed usage example to work with minimum supported ocean version

* added release notes, with a more up to date usage example

joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic if this looks OK to you, let me know and I'll squash it all down for final review ;-)

joseph-wakeling-sociomantic:
... ping? :-P

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,241 @@
...
+        Constructor
+
+        Params:
+            dht = DHT client instance via which to perform the handshake
+            wait_time = time to wait (in seconds) before retrying failed
+                handshake attempts
+
+    ***************************************************************************/
+
+    public this (DhtClient dht, ulong wait_time)
```
nemanja-boric-sociomantic: Nitpick: I like `retry_delay_seconds` (as used in the example) more than `wait_time`.

joseph-wakeling-sociomantic: Sure, can do.  I was copying `wait_time` from `RetryHandshake`, but I agree `retry_delay_seconds` is nicer.

joseph-wakeling-sociomantic:
`fixup` to rename `wait_time` => `retry_delay_seconds`.  Anything else that needs addressing, before I squash?

nemanja-boric-sociomantic:
LGTM. @mihails-strasuns-sociomantic ?

mihails-strasuns-sociomantic (approved):
Very nice!

joseph-wakeling-sociomantic:
Squashed and rebased on latest 13.x.x, and corrected a few small errors in the release-notes usage example:

```diff
diff --git a/relnotes/handshake.feature.md b/relnotes/handshake.feature.md
index 18aef68..610950a 100644
--- a/relnotes/handshake.feature.md
+++ b/relnotes/handshake.feature.md
@@ -8,10 +8,10 @@ applications to support a partial handshake, e.g.:
 
 ```D
 auto retry_delay_seconds = 3;
-handshake = new DhtHandshake(dht_client, retry_delay_seconds);
+auto handshake = new DhtHandshake(dht_client, retry_delay_seconds);
 
 // block on at least one node connecting
-theScheduler.await(this.handshake.oneNodeConnected());
+theScheduler.await(handshake.oneNodeConnected());
 Stdout.formatln("At least one node is now connected!");
 
 // wait until either all nodes have connected, or 60 seconds
@@ -21,7 +21,7 @@ auto timeout_microsec = 60_000_000;
 
 auto handshake_timeout =   // true if timeout is reached
     theScheduler.awaitOrTimeout(
-        this.handshake.allNodesConnected(),
+        handshake.allNodesConnected(),
         timeout_microsec);
 
 if (handshake_timeout)
```

joseph-wakeling-sociomantic:
BTW it would be great if this could be merged early from 13.x.x into 14.x.x, because I'd like to update the imports (`Scheduler` => `IScheduler`) and usage example in that release.

joseph-wakeling-sociomantic:
Oh wait!  Need one last commit-message update ...

joseph-wakeling-sociomantic:
Updated commit message and PR description with cross-ref to https://github.com/sociomantic-tsunami/dhtproto/issues/126.

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,241 @@
...
+        Constructor
+
+        Params:
+            dht = DHT client instance via which to perform the handshake
+            retry_delay_seconds = time to wait (in seconds) before retrying
+                failed handshake attempts
+
+    ***************************************************************************/
+
+    public this (DhtClient dht, ulong retry_delay_seconds)
```
joseph-wakeling-sociomantic: Minor note: `RetryHandshake`'s `wait_time` parameter is `size_t`.  Perhaps we should use `uint` here to guarantee it is compatible?  (Yes, we only support 64-bit, but this is low-hanging fruit to avoid incompatibility with 32-bit.)

mihails-strasuns: We should also use `size_t`.

joseph-wakeling-sociomantic: Why?  It's a weird type to use for a value in seconds.

mihails-strasuns: To match upstream definition and avoid 32-bit incompatibility.

joseph-wakeling-sociomantic: `uint` will be compatible with the `size_t` parameter of `RetryHandshake`, as will `int` with a `verify(retry_delay_seconds > 0)` under the hood (@nemanja-boric-sociomantic expressed a preference for a signed value).

nemanja-boric-sociomantic:
> BTW it would be great if this could be merged early from 13.x.x into 14.x.x, because I'd like to update the imports (Scheduler => IScheduler) and usage example in that release.

I'll leave this decision to @mihails-strasuns-sociomantic, as he was against merging non-tags between the released branches .

mihails-strasuns:
I am fine with merging non-tagged version branches, I was objecting merges in/of major feature branches.

joseph-wakeling-sociomantic:
OK, in the absence of consensus on any other solution, I've updated the constructor parameter to use `size_t` to match `RetryHandshake`:

```diff
diff --git a/src/dhtproto/client/legacy/internal/helper/Handshake.d b/src/dhtproto/client/legacy/internal/helper/Handshake.d
index c924491..e07c900 100644
--- a/src/dhtproto/client/legacy/internal/helper/Handshake.d
+++ b/src/dhtproto/client/legacy/internal/helper/Handshake.d
@@ -131,7 +131,7 @@ public class DhtHandshake
 
     ***************************************************************************/
 
-    public this (DhtClient dht, ulong retry_delay_seconds)
+    public this (DhtClient dht, size_t retry_delay_seconds)
     in
     {
         assert(dht !is null);
```

joseph-wakeling-sociomantic:
... good to go? :-)

mihails-strasuns:
As soon as @gavin-norman-sociomantic is back from vacation ;)

joseph-wakeling-sociomantic:
... _Et exspecto resurrectionem mortuorum..._ ;-)

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -0,0 +1,241 @@
...
+            dht = DHT client instance via which to perform the handshake
+            retry_delay_seconds = time to wait (in seconds) before retrying
+                failed handshake attempts
+
+    ***************************************************************************/
+
+    public this (DhtClient dht, size_t retry_delay_seconds)
+    in
+    {
+        assert(dht !is null);
```
gavin-norman-sociomantic, line 137: Could you check here that the scheduler has really been initialised? (i.e. that the epoll instance to be used is valid.)

gavin-norman-sociomantic (commented):
Apart from the one comment, looks great!

src/dhtproto/client/legacy/internal/helper/Handshake.d
joseph-wakeling-sociomantic: Hmmm, is there a mechanism for this in the ocean version dhtproto 13.x.x needs to support?

More recent ocean versions define an `isSchedulerUsed` function that can be used to check that, but that will only be available from dhtproto 14.x.x onwards.

Note that `theScheduler` has an out-contract that verifies that the underlying `Scheduler` instance has been initialized (i.e. it can't return `null`, and therefore in turn `initScheduler` must have been called).

gavin-norman-sociomantic: Oh that's fine then :+1:.

joseph-wakeling-sociomantic: Cool!  Let me know once it's merged into 14.x.x: it may be worth tweaking it to use `IScheduler`, and we _can_ use `isSchedulerUsed()` as an extra check if wanted.

PR #134, opened 2018-04-09
Use IScheduler and awaitOrTimeout in DhtHandshake implementation and usage example

joseph-wakeling-sociomantic:
We couldn't do this in https://github.com/sociomantic-tsunami/dhtproto/pull/129, since dhtproto v13.x.x needs to support older ocean versions that do not have the `IScheduler` interface or the `Scheduler.awaitOrTimeout` method.

### DIFF
--- a/src/dhtproto/client/legacy/internal/helper/Handshake.d
+++ b/src/dhtproto/client/legacy/internal/helper/Handshake.d
@@ -215,27 +215,37 @@ unittest
             theScheduler.await(this.handshake.oneNodeConnected());
             Stdout.formatln("At least one node is now connected!");
 
-            // if `theScheduler.awaitOrTimeout` is available, it can
-            // be used to implement support for a partial handshake:
+            // wait until either all nodes have connected, or 60 seconds
+            // have passed, whichever comes sooner
             //
-            // ```
-            // auto timeout_microsec = 60_000_000;
-            //
-            // auto handshake_timed_out =   /* true if timeout is reached */
-            //     theScheduler.awaitOrTimeout(
-            //         this.handshake.allNodesConnected(),
-            //         timeout_microsec);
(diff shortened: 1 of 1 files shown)


joseph-wakeling-sociomantic:
@gavin-norman-sociomantic just to note, it would be trivial to also add a `verify(isSchedulerUsed())` line to the constructor.  I haven't done this on the vaguely remembered notion that it's considered Bad Form to throw exceptions from a constructor.

src/dhtproto/client/legacy/internal/helper/Handshake.d
```diff
@@ -215,27 +215,37 @@ unittest
...
-            //    that the `DhtHandshake` instance will keep working
-            //    in the background to complete the handshake */
-            // ```
-            //
-            // alternatively, we can block on all DHT nodes being
-            // connected using `await`:
-            theScheduler.await(this.handshake.allNodesConnected());
-
-            Stdout.formatln("DHT handshake complete!");
+            // (if you really need a hard block until the full handshake
```
gavin-norman-sociomantic, line 221: If you wanted that behaviour, there'd be no reason to call `oneNodeConnected` at all, right?

joseph-wakeling-sociomantic: Well, it depends.  In apps I'm responsible for, I'll probably make it a configurable option whether to `await` or `awaitOrTimeout` on all nodes being connected.  So the separation still has some meaning.

gavin-norman-sociomantic: Fair enough. The comment is innocent enough. I just wondered.

gavin-norman-sociomantic (approved):

PR #136, opened 2018-04-13
Add built-in logging to RetryHandshake

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Ahh needs release notes.

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
```diff
@@ -118,10 +86,11 @@ class RetryHandshake
 
     /***************************************************************************
 
-        try doing the handshake
+        Try doing the handshake.
```
joseph-wakeling-sociomantic, line 100: Minor style: I thought we didn't have `.` at the end of the 'title' line of a ddoc block?

gavin-norman-sociomantic: Hmm, I always add one. And it's purely a coincidence that this comment is just one line long.

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
```diff
@@ -25,73 +25,41 @@ class RetryHandshake
...
 
-    /***************************************************************************
-
-        Delegate that will be called on the first successful handshake with each
-        individual node
-
-    ***************************************************************************/
-
+    /// Delegate that will be called on the first successful handshake with each
+    ///individual node.
```
joseph-wakeling-sociomantic: Minor style: space after `///` ... ?

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
joseph-wakeling-sociomantic: That's weird, I have had the impression for a long time that the absence of one was actually a style requirement :-P

gavin-norman-sociomantic: I've never heard of it, if it is! If you can find a ref to this in our style guide, I'm happy to rethink my existence!

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
```diff
@@ -203,6 +182,8 @@ class RetryHandshake
             {
                 this.established_nodes[node_hash] = true;
                 this.one_node_handshake_dg(info.nodeitem);
+                this.log.info("Handshake succeeded on {}:{}.",
```
joseph-wakeling-sociomantic, line 188: This is going to be very noisy in production uses.  Are you sure it should be `info` and not `trace` ... ?

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
joseph-wakeling-sociomantic: Well, the style guide has a `.`, so maybe I should rethink mine! ;-)

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
gavin-norman-sociomantic: Hmm good question. I figured that as this happens only once ever per node, it's not too bad. And it's pretty useful logging.

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
gavin-norman-sociomantic: Updated.

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
joseph-wakeling-sociomantic: Yup, on balance I agree.  It's straightforward enough to redirect the logging for this module to another file (this is useful with e.g. the new `Mirror` class, where the entries are not just once on app startup).

gavin-norman-sociomantic: That's good. (I'd wondered about that, as the logger name is determined at run-time.)

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
```diff
@@ -113,32 +86,34 @@ class RetryHandshake
 
         this.timer = new TimerEvent(&this.tryHandshake);
 
+        this.log = Log.lookup(format("RetryHandshake instance 0x{:x}",
```
gavin-norman-sociomantic: Do you think there's any point to formatting the logger name like this? Theoretically, a user might have >1 DHT client, each with their own `RetryHandshake` instance.

joseph-wakeling-sociomantic: I don't mind one way or the other, but it might be nicer to allow the user to specify a custom identifier if that is the intended use-case.

nemanja-boric-sociomantic: I suggest having structured logger name:

```
RetryHandshake.Instance 0x{:x}
```

gavin-norman-sociomantic: Why use a `.` between the first two words but not the second?

nemanja-boric-sociomantic: I think `RetryHandshake.0x{:x}` also works. The trick is that the config file could then do `LOG.RetryHandshake.*`

nemanja-boric-sociomantic: (iow, I'm not sure if the globbing works on the arbitrary point in the logger name)

gavin-norman-sociomantic: Oh is `.` a magic character for the logger framework? (I wasn't sure if it is or not.)

gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

joseph-wakeling-sociomantic (approved):
LGTM.  I would still consider adding the possibility of a user-specified identifier in future, rather than just using the pointer value.

gavin-norman-sociomantic:
>I would still consider adding the possibility of a user-specified identifier in future, rather than just using the pointer value.

Yeah... It's a "just in case" feature, really, rather than something that's intended to be useful. I was really on the edge about including it at all :/

joseph-wakeling-sociomantic:
TBH in that case, I would drop it for now, and consider adding identifier support in future if there's a use-case for it.

gavin-norman-sociomantic:
That sounds reasonable.

joseph-wakeling-sociomantic:
(As a minor note: the identifier that matters is surely that of the DHT client rather than the `RetryHandshake` instance?)

gavin-norman-sociomantic:
I'd make it a module level logger, in that case.

joseph-wakeling-sociomantic:
> I'd make it a module level logger, in that case.

I wouldn't.  It makes things more flexible and encapsulated to just keep it inside the class.

gavin-norman-sociomantic:
Encapsulated, I can see. But how is it more flexible?

nemanja-boric-sociomantic:
I really doubt any usefulness of the pointer value, except that it's different from the other value :-). (So it doesn't really matter if it's the address of the client or retry handshake object).

gavin-norman-sociomantic:
>I really doubt any usefulness of the pointer value, except that it's different from the other value :-).

Right, that's the point. _Just in case_ anyone had two instances of this class, the log messages could be distinguished from each other.

joseph-wakeling-sociomantic:
> Encapsulated, I can see. But how is it more flexible?

Because if in future you do want to add some unique identifier, you don't have to rewrite it as a class parameter.

gavin-norman-sociomantic:
Updated with per-instance ref to shared logger.

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
```diff
@@ -24,74 +24,48 @@ class RetryHandshake
...
+    import ocean.util.log.Logger;
 
-    /***************************************************************************
-
-        Timer to retry the handshake
-
-    ***************************************************************************/
+    /// Logger for instances of this class.
+    /// (Note that this assumes there is only one instance of this class. This
+    /// is the case in practice. If we ever TODO TODO)
```
joseph-wakeling-sociomantic: `TODO TODO`?!

gavin-norman-sociomantic: Oops. That was obviously meant to be a message from my Friday self to my self now. I completely missed it :D

gavin-norman-sociomantic: Updated.

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
```diff
@@ -113,32 +90,33 @@ class RetryHandshake
 
         this.timer = new TimerEvent(&this.tryHandshake);
 
+        this.log = Log.lookup("RetryHandshake");
```
joseph-wakeling-sociomantic, line 93: Minor: `DhtHandshake` goes with the full module `dhtproto.client.legacy.internal.helper` prefix.  Should we aim for consistency here?

It doesn't really matter which way the consistency goes, but it might be a good idea for the `RetryHandshake` logging to be clear it's part of `dhtproto`.

joseph-wakeling-sociomantic (commented):
Looks all right to me, although I have one small query about the logging.

I was going to say something about the `static Logger`, but you got there first ;-)

src/dhtproto/client/legacy/internal/helper/RetryHandshake.d
gavin-norman-sociomantic: Hmm, my tendency (not that I've done this very often) is to only give full module path names to loggers that are instantiated at module level. Not sure if that's an important distinction or not.

For reference, the logger in the new channel mirror does not have the full path name.

joseph-wakeling-sociomantic: It seems a reasonable enough distinction; it'd just be nice to be unambiguous that this is DHT-related.  But certainly not a big deal enough to keep reworking this.

For the future, on identifiers in logs: my inclination here is that it makes sense that any identifier be associated with the client instance, and helper classes like this reference the client they are working with.  That would make it easy for other wrappers (e.g. `DhtHandshake`) to incorporate these identifiers without needing to coordinate between each other.

PR #137, opened 2018-04-17
Convert assert -> verify

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Updated with change to `handleNone()`.

PR #138, opened 2018-04-18
Add Mirror logging when a request first starts receiving data

gavin-norman-sociomantic:


gavin-norman-sociomantic:
@joseph-wakeling-sociomantic how's this look? Could you try it out, if it looks helpful?

joseph-wakeling-sociomantic:
OK, just gave this a go.  2 points of concern:

* it means that a `Started receiving data` shows up for every recurring `GetAll` from the mirrors

* I'm not sure it's really what is wanted, which is to have a clear indicator that a previously-failing request has succeeded on the next retry

joseph-wakeling-sociomantic:
I think it's OK in itself, though, so as a "better than nothing" option, LGTM.

src/dhtproto/util/Verify.d
```diff
@@ -0,0 +1,80 @@
...
+    be safely merged into v14.
+
+    Copyright: Copyright (c) 2018 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module dhtproto.util.Verify;
```
joseph-wakeling-sociomantic, line 15: If it requires adding a whole module as a workaround, I wonder if it might be preferable to just target this feature against dhtproto 14.x.x.

joseph-wakeling-sociomantic (approved):
All things considered this seems better to have than not.  But you might want to consider dropping `dhtproto.util.Verify` and just retargeting this against dhtproto v14.x.x.

src/dhtproto/util/Verify.d
gavin-norman-sociomantic: It's likely this will be required by any changes to v13.x.x, from now on, so adding it isn't just for the benefit of this specific PR.

nemanja-boric-sociomantic (approved):
I'll open the issue to get rid of the `Verify` at least in v15.0.0

PR #161, opened 2018-07-11
Replace `getMsg` with `message`

gavin-norman-sociomantic:


mihails-strasuns:
Haven't we agreed on just releasing v13.x.x right away to avoid duplicated commits?

gavin-norman-sociomantic:
Yes, but then I subsequently changed my mind. There's nothing of note unreleased in v13.x.x, so making a minor release doesn't really make sense. I'd prefer to do this as a patch release, as originally planned.

mihails-strasuns-sociomantic:
Okay, as you prefer.

Issue #82, opened 2017-12-20
Partial handshake support: Mirror disk dump / load helper


gavin-norman-sociomantic:
Some applications have code that does the following on a DHT channel:
* Uses the legacy `ChannelMirror` to mirror all records.
* Stores all received records in a cache or map in memory.
* Periodically / asynchronously writes the mirrored channel to disk.
* At shutdown, saves the mirrored channel to disk.
* At startup, loads the mirrored channel from disk then starts the channel mirror.

This sounds like useful behaviour for many apps and would be nice to have in a core library somewhere (using the neo Mirror request instead of `ChannelMirror`, of course). The question is how to break it down into one or more components that would sit nicely in the existing libraries.

gavin-norman-sociomantic:
So this sounds like this could be broken down into 3 components, to me:
1. A DHT Mirror request. (That already encapsulates all the complexity that was managed by the legacy `ChannelMirror` class.)
2. An ocean cache enhanced with a periodic dump (probably somewhat abstract, not bound to any particular destination).
3. A means of asynchronous dumping. e.g. `SafeFork` or the [async disk I/O framework from the DLS node](https://github.com/sociomantic-tsunami/dlsnode/tree/master/src/dlsnode/util/aio).

nemanja-boric-sociomantic:
Note that `SafeFork` is only half-safe: https://github.com/sociomantic-tsunami/ocean/issues/315

However, this can (and should and will) be fixed, and the good thing about forking is that it effectively does all critical section behaviour you need via copy-on-write mechanism. Because, in order to serialize the data structure, you need to make sure that changing it while dumping it to disk don't mess up with the data that already made it to the file, which is not a trivial problem for most data structures.

Alternative to all this is using disk persistent data structures, such as `BTree` and asynchronous dumping, but that sounds more like an enhancement to the existing cache/fork approach.

gavin-norman-sociomantic:
I guess it's two different approaches:
1. Periodically fork and do a complete dump.
2. Write cache updates to disk as they come in (async I/O).

I'd be more inclined to favour the latter, but we can assess pros / cons.

gavin-norman-sociomantic:
Alligator's implementation: https://github.com/sociomantic/alligator/blob/953425dcb1b56b711a8d8ceffcf96d3cea0e238c/src/alligator/cache/Cache.d.

joseph-wakeling-sociomantic:
[...]

PR #162, opened 2018-07-13
Mirror disk dump / load helper

gavin-norman-sociomantic:
Rough concept. Open to all feedback as to the API / implementation / usefulness.

### DIFF
--- a/src/dhtproto/client/helper/ChannelSerializer.d
+++ b/src/dhtproto/client/helper/ChannelSerializer.d
@@ -1,0 +1,373 @@
+/*******************************************************************************
+
+    DHT channel serializer helper.
+
+    Helper class template for dumping a cached or mirrored channel from a map
+    (of some kind) in memory to disk, then loading it again at startup. The
+    dumping can be done in either a blocking way (e.g. at application shutdown)
+    or asynchronously via `fork` (e.g. periodically while the application is
+    running).
+
+    Notes:
+        * In-memory data structure: Is assumed to be some kind of map, from
+          `hash_t` keys to `Contiguous!(S)` values (where `S` is the type of
+          the deserialized struct).
+        * Blocking dump: Dumps the map without forking. If an asynchronous dump
+          is already in progress, the call will block until the dump is
+          finished, then start it again.
+        * Asynchronous dump: When an asynchronous dump is in progress, further
+          calls to `dump` will not start a second, clashing dump; the dump
+          method will simply return.
+
+    Copyright:
+        Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module dhtproto.client.helper.ChannelSerializer;
+
+import ocean.transition;
+import ocean.io.device.File;
+import ocean.io.serialize.SimpleStreamSerializer;
+import ocean.io.stream.Buffered;
+import ocean.sys.SafeFork;
+import ocean.util.container.map.Map;
+import ocean.util.serialize.contiguous.Contiguous;
+import ocean.util.serialize.contiguous.Deserializer;
+import ocean.util.serialize.contiguous.MultiVersionDecorator;
+import ocean.util.serialize.contiguous.Serializer;
+import ocean.util.serialize.Version;
+
+/*******************************************************************************
+
+    DHT channel serializer helper.
+
+    Params:
+        S = type of record value stored in cached / mirrored channel
+
+*******************************************************************************/
+
+public class ChannelSerializer ( S )
+{
+    /// Path of the dump file.
+    private istring path;
+
+    /// Buffered output instance.
+    private BufferedOutput buffered_output;
+
+    /// Buffered input instance.
+    private BufferedInput buffered_input;
+
+    /// Fork helper for asynchronous dumping.
+    private SafeFork fork;
+
+    /// Delegate type required by `dump`.
+    private alias void delegate ( void delegate ( hash_t, ref Contiguous!(S) ) )
+        DumpRecordDg;
+
+    /// Delegate type required by `load`.
+    private alias void delegate ( hash_t, ref Contiguous!(S) ) LoadRecordDg;
+
+    /// Delegate used by the fork for providing the records to be dumped. (Note
+    /// that the fork copy-on-write behaviour means it's safe for this value to
+    /// be changed while a fork is running.)
+    private DumpRecordDg record_dump_dg;
+
+    /// Buffer used for de/serialization of record values.
+    private void[] serialize_buf;
+
+    /// Version decorator, for record value structs that have version info.
+    static if ( Version.Info!(S).exists )
+        private VersionDecorator version_decorator;
+
+    /***************************************************************************
+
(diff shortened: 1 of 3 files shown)


src/dhtproto/client/helper/ChannelSerializer.d
```diff
@@ -0,0 +1,354 @@
...
+
+    /// Buffered input instance.
+    private BufferedInput buffered_input;
+
+    /// Fork helper for asynchronous dumping.
+    private SafeFork fork;
+
+    /// TODO
+    private alias void delegate ( void delegate ( ref hash_t, ref S ) )
+        DumpRecordDg;
```
joseph-wakeling-sociomantic, line 69: Might be worth supporting `Contiguous!(S)` as dump input as well, since that may well be how the data is received and stored by the application.

gavin-norman-sociomantic: Ah yes, good point. That ought to be the standard API.

src/dhtproto/client/helper/ChannelSerializer.d
```diff
@@ -0,0 +1,354 @@
...
+            aa = associative array to dump
+            block = if true, performs a blocking dump; otherwise performs an
+                asynchronous dump (see module header)
+
+        Returns:
+            true if a dump occurred; false if it was skipped
+
+    ***************************************************************************/
+
+    public bool dump ( HashMap!(S) map, bool block = false )
```
joseph-wakeling-sociomantic: Bear in mind that applications may be using other cache types than built-in AA's or `HashMap` (e.g. `StandardKeyKashingMap`).

gavin-norman-sociomantic: Ah yes, I was thinking that `StandardKeyKashingMap` was dervied from `HashMap`, but it's not.

Arbitrary types are supported too, by the last overload.

joseph-wakeling-sociomantic: Gotcha, makes sense.  Although it might be possible to simplify the different overloads somewhat by implementing a generic mechanism that supports any type that can be `foreach`'d over.

gavin-norman-sociomantic: The map and AA overloads are purely convenience wrappers around the overloads for generic types.

joseph-wakeling-sociomantic: That said in some ways I feel it might be simpler for the helper to avoid any need to care about the actual cache, and just provide the means to dump individual key/value pairs.

gavin-norman-sociomantic: Yeah, the trouble is the dumping of an individual record has to occur inside `SafeFork`, so it's not something that can be called via the public API. The user has to register their intention to dump, then the `SafeFork` instance asks them for the records to be dumped, when it's ready.

gavin-norman-sociomantic:
@scott-gibson-sociomantic could you or someone from your team have a look at this? It's designed for your applications that need a complete mirror of a DHT channel at start up.

gavin-norman-sociomantic:
Rebased onto v14.x.x. (v13.x.x is out of support.)

src/dhtproto/client/helper/ChannelSerializer.d
```diff
@@ -0,0 +1,382 @@
...
+        Params:
+            aa = associative array to load into
+
+    ***************************************************************************/
+
+    public void load ( ref Contiguous!(S)[hash_t] aa )
+    {
+        void loadRecord ( hash_t k, ref Contiguous!(S) v )
+        {
+            aa[k] = v;
```
andrej-mitrovic-sociomantic, line 190: Is there any memory management you have to do here? For example, when loading keys/values which have arrays, would you have to manually duplicated them, or is this already handled somewhere else?

andrej-mitrovic-sociomantic: In Alligator we used to have to do this manually, because we used the MapSerializer which uses a buffer for the keys / values, and there we had to copy the arrays manually.

gavin-norman-sociomantic: I believe it's handled by the deserialization code below (https://github.com/sociomantic-tsunami/dhtproto/pull/162/commits/4a31c38d3e3af1f71b07b69741992af9c9394137#diff-5c38ebc93bed0b168f5f68fe43193a84R240), but we should definitely confirm that.

gavin-norman-sociomantic: Yep. Looks like `v` (https://github.com/sociomantic-tsunami/dhtproto/pull/162/commits/4a31c38d3e3af1f71b07b69741992af9c9394137#diff-5c38ebc93bed0b168f5f68fe43193a84R236) will initially contain an empty buffer, which will be allocated and filled with the deserialized struct. The reference to this is then copied into the AA or map.

gavin-norman-sociomantic: This is one of the nice things about working with `Contiguous!(S)`, by the way.

andrej-mitrovic-sociomantic: Nice.

gavin-norman-sociomantic:
Finalised and unblocked.

gavin-norman-sociomantic:
As this PR implements a helper that is needed for some apps to support partial DHT handshake effectively, I feel we should merge it without application testing. When we come to adapt some app to use it, we can iron out any issues that come up.

src/dhtproto/client/helper/ChannelSerializer.d
```diff
@@ -0,0 +1,381 @@
...
+    /***************************************************************************
+
+        SafeFork entry point. Calls `this.new_record_dump_dg` to dump the
+        provided data structure to disk.
+
+    ***************************************************************************/
+
+    private void forkEntryPoint ( )
+    {
+        this.current_record_dump_dg = this.new_record_dump_dg;
```
matthias-wende-sociomantic: This is not needed due to the fork.

matthias-wende-sociomantic (commented):
Looks good. Just one comment.

gavin-norman-sociomantic:
I just noticed this PR doesn't implement the second feature mentioned in https://github.com/sociomantic-tsunami/dhtproto/issues/82#issuecomment-380724510.

>A means of handling errors while writing.

Example solutions:
1. Make a backup of the old dump before writing the new one, and load the backup if an error occurs while loading the dump.
2. Use the approach used in the dhtnode: write dumps to a temp file and rename it on success.

matthias-wende-sociomantic (approved):

scott-gibson-sociomantic:
I just realised, this PR uses no other swarm/dhtproto imports? Why does this live in dhtproto?

matthias-wende-sociomantic:
We talked about that and yes it is totally independent :). AFAIR the reason is that the use case of this helper is mainly restricted to the DHT. If someone comes up with a different use case independent of dhtproto it might be worth to consider moving it to ocean.

gavin-norman-sociomantic:
Exactly. The plan was/is also to enhance this to integrate it with the DHT client / channel mirror. (Though this module may remain independent.)

PR #178, opened 2018-08-28
Fix client's blocking hash-range query helpers

### LINKED ISSUE #177: HashRangesKnown task hangs despite all nodes connecting
The `HashRangesKnown` task is supposed to block until at least the specified number of nodes are connected.  However, when trying this out with an app being ported to neo, I found that it seems to block indefinitely despite all nodes being connected.

The code below outlines how the task is used: in order to support operating with partial handshake, the app first awaits completion of a `HashRangesKnown(1)` instance (ensuring that at least one node hash range has been obtained), before proceeding with an `awaitOrTimeout` on all hash ranges being known.

```D
// wait for dht.neo protocol to obtain all hash node ranges
log.info("Waiting for DHT.neo nodes hash ranges ...");
theScheduler.await(dht.blocking.new HashRangesKnown(1));

log.info("At least one DHT.neo hash range known");

bool neo_handshake_timed_out = false;
auto neo_timeout_sec = 60;

if (neo_timeout_sec > 0)
{
    neo_handshake_timed_out =
        theScheduler.awaitOrTimeout(
            dht.blocking.new AllHashRangesKnown,
            neo_timeout_sec * 1_000_000 /* microsec */);
}
else
{
    dht.blocking.waitAllHashRangesKnown();
}

if (neo_handshake_timed_out)
{
    log.error("DHT.neo hash ranges not complete within "
              ~ "{} seconds.", neo_timeout_sec);
}
else
{
    log.info("All DHT.neo hash ranges known.");
}
```

When run, the DHT client logs seem to report connections to both the local test nodes used, but the "At least one" log message is never output and the app logic does not proceed beyond the point where `theScheduler.await` is called with the `HashRangesKnown` instance:

```
Info [...] Waiting for DHT.neo nodes hash ranges ...
Trace [...] <connected>: Node [127, 0, 0, 1]:9100
Trace [...] <connected>: Node [127, 0, 0, 1]:9101
Trace [...] <hash_range_queried>: Node [127, 0, 0, 1]:9100
Trace [...] <hash_range_queried>: Node [127, 0, 0, 1]:9101
```

gavin-norman-sociomantic:
The check for the finishing condition was previously triggered using
swarm's waitConnect(). This method only triggers upon connection events.
However, to implement the correct behaviour for the DHT client's hash-range
query helpers, we need to trigger upon hash-range query events, not
connection events.

Fixes #177.

### DIFF
--- a/src/dhtproto/client/mixins/NeoSupport.d
+++ b/src/dhtproto/client/mixins/NeoSupport.d
@@ -614,39 +614,15 @@ template NeoSupport ( )
 
         public void waitAllHashRangesKnown ( )
         {
-            auto task = Task.getThis();
-            verify(task !is null, "This method may only be called from inside a Task");
-
-            auto old_user_conn_notifier = this.outer.user_conn_notifier;
-
-            scope conn_notifier =
-                ( Neo.DhtConnNotification info )
-                {
-                    old_user_conn_notifier(info);
-
-                    with ( info.Active ) switch ( info.active )
-                    {
-                        case hash_range_queried:
-                            if ( task.suspended() )
-                                task.resume();
-                            break;
-
-                        case connected:
-                        case connection_error:
-                            break;
-
-                        default: assert(false);
-                    }
-                };
-
-            this.outer.user_conn_notifier = conn_notifier;
-            scope ( exit )
-                this.outer.user_conn_notifier = old_user_conn_notifier;
-
             scope stats = this.outer.neo.new DhtStats;
 
-            while ( stats.num_nodes_known_hash_range < stats.num_registered_nodes )
-                task.suspend();
+            bool finished ( )
+            {
+                return stats.num_nodes_known_hash_range ==
+                    stats.num_registered_nodes;
+            }
+
+            this.waitHashRangeQuery(&finished);
         }
 
         /***********************************************************************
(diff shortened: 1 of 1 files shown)


joseph-wakeling-sociomantic:
@gavin-norman-sociomantic this seems to work, thanks for the quick fix :-)

gavin-norman-sociomantic:
Excellent. I'll get @nemanja-boric-sociomantic to review it when he's back.

src/dhtproto/client/mixins/NeoSupport.d
```diff
@@ -721,12 +697,60 @@ template NeoSupport ( )
...
+        /***********************************************************************
+
+            Suspends the current task until the specified finished condition is
+            satisifed.
+
+            Params:
+                finished = delegate specifying the condition under which the
+                    method will return. The delegate is called each time the
+                    connection notifier is called (i.e. on connection and hash-
+                    range query events)
```
joseph-wakeling-sociomantic: This doc appears a little misleading: AFAICS the delegate is not called each time the connection notifier is called, but rather once on the initial function call and thereafter (unless we return) every time a hash-range query is received.

gavin-norman-sociomantic: You're right. Good catch.

gavin-norman-sociomantic: Updated.

joseph-wakeling-sociomantic:
From a documentation point of view LGTM.  Technically looks OK too but I'll leave it to the experts to give a final view on that ;-)

david-eckardt-sociomantic:
LGTM. To be sure, @nemanja-boric-sociomantic should take a last look.

nemanja-boric-sociomantic (approved):

Issue #174, opened 2018-08-24
Legacy client connection availability tracker


gavin-norman-sociomantic:
Some applications have a data flow pattern like the following:
1. Receive a record from an incoming source (e.g. a DMQ channel).
2. Do some expensive processing on the record. (e.g. querying an external service.)
3. Write the results of the processing to the DHT.

In the case where 2 is expensive but non-critical, it is sometimes preferable to not perform step 2 at all, if the DHT node to which the result would be written is inaccessible. (i.e. rather than queuing up the write to be performed when the DHT node becomes accessible.)

The trick here is that there's no built-in way to check if a connection is up or not. We could add this feature in the neo client easily enough, but for the legacy client it's trickier.

As this feature is required in multiple apps, the suggestion is to add a helper to dhtproto that does the following:
* Owns a data structure that stores a flag per node in the client's registry, indicating whether it recently had an error or not. This should be accompanied by a timestamp, indicating the last time an error was logged.
* Has a method to be called from DHT request notifiers, if an I/O error is indicated. Sets the flag for the appropriate node and set the timestamp to now.
* Has a method `isNodeAvailable`:
  * If the error flag for the node is false, then the node is available.
  * If the error flag is true and the timestamp of the last error is "some time ago" (how this is defined tbd), then set the flag to false and return that the node is available.

PR #184, opened 2018-10-16
Connection availability tracker

gavin-norman-sociomantic:
Addresses #174.

### DIFF
--- a/src/dhtproto/client/legacy/internal/helper/NodeAvailability.d
+++ b/src/dhtproto/client/legacy/internal/helper/NodeAvailability.d
@@ -1,0 +1,195 @@
+/*******************************************************************************
+
+    Tracker for availability of nodes, based on recency of connection errors.
+
+    This helper is useful in apps that have a data flow pattern like the
+    following:
+        1. Receive a record from an incoming source (e.g. a DMQ channel).
+        2. Do some expensive but non-critical processing on the record. (e.g.
+           querying an external service.)
+        3. Write the results of the processing to the DHT.
+
+    By tracking which DHT nodes have had connectivity problems recently, the app
+    can decide to not perform step 2 at all, if the DHT node to which the result
+    would be written is inaccessible. (i.e. discarding the request, rather than
+    queuing up the write to be performed when the DHT node becomes accessible.)
+
+    Copyright:
+        Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved.
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module dhtproto.client.legacy.internal.helper.NodeAvailability;
+
+/// ditto
+public struct NodeAvailabilty
+{
+    import swarm.Const;
+    import core.sys.posix.time;
+    import dhtproto.client.legacy.internal.registry.model.IDhtNodeRegistryInfo;
+
+    /// The number of seconds for which the `available()` methods will report a
+    /// node as inaccessible after an error is reported for it.
+    public uint retry_delay_s = 3;
+
+    /***************************************************************************
+
+        Should be called when a request to the specified node fails due to a
+        connection error.
+
+        Params:
+            node = node that had a connection error
+
+    ***************************************************************************/
+
+    public void error ( NodeItem node )
+    {
+        if ( auto tracker = node.toHash() in this.error_trackers )
+        {
+            tracker.error();
+        }
+        else
+        {
+            auto tracker = NodeErrorTracker(this);
+            tracker.error();
+            this.error_trackers[node.toHash()] = tracker;
+        }
+    }
+
+    /***************************************************************************
+
+        Tells whether the node that is responsible for the specified key is
+        believed to be currently accessible or not.
+
+        Params:
+            key = key of record to be accessed
+            registry = node registry to look up which node is responsible for
+                the specified key
+
+        Returns:
+            true if the node is believed to be accessible; false if it has had a
+            connection error recently
+
+    ***************************************************************************/
+
+    public bool available ( hash_t key, IDhtNodeRegistryInfo registry )
+    {
+        auto node = registry.responsibleNode(key);
+        if ( node is null )
+            return false;
+
+        NodeItem nodeitem;
+        nodeitem.Address = node.address;
+        nodeitem.Port = node.port;
+
+        return available(nodeitem);
+    }
+
+    /***************************************************************************
+
+        Tells whether the specified node is believed to be currently accessible
+        or not.
+
+        Params:
+            node = node to be accessed
+
+        Returns:
+            true if the node is believed to be accessible; false if it has had a
+            connection error recently
+
+    ***************************************************************************/
+
+    public bool available ( NodeItem node )
+    {
+        if ( auto tracker = node.toHash() in this.error_trackers )
+            return tracker.available();
+        else
+            return true;
+    }
+
+    /// Internal error tracker for a single node.
+    private struct NodeErrorTracker
+    {
+        /// Pointer to the outer instance.
+        NodeAvailabilty* outer;
+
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic:
Rough, initial implementation of #174. As potential users, maybe could @vasileios-kyrillidis-sociomantic @matthias-wende-sociomantic have a look?

src/dhtproto/client/legacy/internal/helper/NodeAvailability.d
```diff
@@ -0,0 +1,82 @@
...
+        public void error ( )
+        {
+            this.last_error_timestamp = time(null);
+        }
+
+        /// TODO
+        public bool available ( )
+        {
+            const delay_ms = 3_000; // TODO: configurable
+            auto diff = time(null) - this.last_error_timestamp;
```
nemanja-boric-sociomantic: We should use monotonic clock here which is not prone to adjustments.

src/dhtproto/client/legacy/internal/helper/NodeAvailability.d
```diff
@@ -0,0 +1,82 @@
...
+    /// TODO
+    private struct NodeErrorTracker
+    {
+        /// TODO
+        private time_t last_error_timestamp;
+
+        /// TODO
+        public void error ( )
+        {
+            this.last_error_timestamp = time(null);
```
nemanja-boric-sociomantic: ditto monotonic clock.

vasileios-kyrillidis-sociomantic (commented):
Thankfully Nemanja already spotted something I wouldn't :D
Apart from that it looks good to me! 

A bit offtopic: In general I prefer builder type objects to structs when they contain container instances (hash map in this case), but this is a perfectly acceptable solution so it LGTM!

gavin-norman-sociomantic:
>A bit offtopic: In general I prefer builder type objects to structs when they contain container instances (hash map in this case), but this is a perfectly acceptable solution so it LGTM!

What would a "builder type object" look like in this case?

gavin-norman-sociomantic:
Updated with monotonic clock.

gavin-norman-sociomantic:
@vasileios-kyrillidis-sociomantic also note that, as I suspected, it's simpler to track when an error happened, rather than when an error _didn't_ happen (like in your older implementation).

src/dhtproto/client/legacy/internal/helper/NodeAvailability.d
```diff
@@ -0,0 +1,89 @@
...
+        {
+            this.last_error_timestamp = time();
+        }
+
+        /// TODO
+        public bool available ( )
+        {
+            const delay_ms = 3_000; // TODO: configurable
+            auto diff = time() - this.last_error_timestamp;
+            return diff >= delay_ms;
```
nemanja-boric-sociomantic: You're comparing seconds with milliseconds here.

I suggest you change `time` to return the time in milliseconds:

`return 1000*ts.tv_sec + ts.tv_nsec / 1_000_000`.

gavin-norman-sociomantic: Haha yes. I've not tested this yet ;)

gavin-norman-sociomantic:
Updated with doc and (small) unittest.

@vasileios-kyrillidis-sociomantic would you be able to test this in your app? Trying it out in a real world app would be the best way of checking it.

vasileios-kyrillidis-sociomantic:
Sure I ll test this hopefully tomorrow.

gavin-norman-sociomantic:
Thanks! Blocking until tested.

vasileios-kyrillidis-sociomantic:
> What would a "builder type object" look like in this case?

A class with setters. Pretty much the same, just with `.setNodes()` method instead.

gavin-norman-sociomantic:
>A class with setters. Pretty much the same, just with .setNodes() method instead.

Mm, you mean instead of passing the nodes registry to the `available` method? https://github.com/sociomantic-tsunami/dhtproto/pull/184/files#diff-274c306349464b80c856d501990f38e7R78

vasileios-kyrillidis-sociomantic:
Yes, I was thinking something like you set the nodes at some point in the beginning and then just use the availability methods. But as said, it's not a huge difference.

gavin-norman-sociomantic:
I see. Yeah, it could be a slightly more convenient API. I think it doesn't make much difference, though, as wherever you use the connection tracker you by definition have access to the DHT client instance, which contains the nodes registry.

vasileios-kyrillidis-sociomantic:
Sorry ended up not looking into this today. I promise I will on Monday :)

vasileios-kyrillidis-sociomantic:
I have just finished testing this on a local DHT with 2 DHT nodes and it seems to work fine. 

Only thing I realized is that I would like to have an `error` method with a signature like the equivalent available method `public void error (hash_t key, IDhtNodeRegistryInfo registry)`.

At least in Booty, we mark a node to be "down" when a DHT Put fails, meaning that I have the record key at hand and would be nice to directly "query" the node item internally in NoveAvailability (as done in the available method).

gavin-norman-sociomantic:
>At least in Booty, we mark a node to be "down" when a DHT Put fails, meaning that I have the record key at hand and would be nice to directly "query" the node item internally in NoveAvailability (as done in the available method).

Can't you use the `NodeItem` struct in the notification? https://github.com/sociomantic-tsunami/swarm/blob/v5.x.x/src/swarm/client/request/notifier/IRequestNotification.d#L134

vasileios-kyrillidis-sociomantic:
Of course it can be done. It's just for 2 minor reasons I wrote the above:
1) It's just that in the current logic we time a request out after a small amount of time, meaning outside of the notifier. I would have to store a reference to the node item and use it outside (so no real problem here). 
2) I would expect the API to be similar for both "error" and "available"

gavin-norman-sociomantic:
>It's just that in the current logic we time a request out after a small amount of time, meaning outside of the notifier.

Hm, can you explain this a bit more? Doesn't a request timeout also call the notifier? (Perhaps I'm forgetting some aspect of the legacy DHT client.)

>I would expect the API to be similar for both "error" and "available"

They were intended to be used from completely different contexts (`available` before assigning a request and `error` from inside a request's notifier), hence the different signatures.

vasileios-kyrillidis-sociomantic:
> Doesn't a request timeout also call the notifier? (Perhaps I'm forgetting some aspect of the legacy DHT client.)

No not really, the code looks like this: 
```
    StopWatch stop_watch;
    stop_watch.start();
    while (!assignPutRequest(put_request)) // Notifier is in the assignPutRequest
    {
        // Check if we should time this request out
        // NOTE: It's ok to lose some precision here by casting to uint
        uint elapsed_millis = cast(uint)(stop_watch.microsec / 1000);
        if (elapsed_millis >= timeout_wait_millis)
        {
            auto node_item = getResponsibleDhtNodeItem(key);
            node_availability.error(node_item);
            return false;
        }

        // Otherwise wait for a bit and then retry
        wait(dht_config.retry_wait_millis);
    }
```

Edit: Basically we manually track the timeouts.

gavin-norman-sociomantic:
That's... extremely unusual usage! Why are you doing it that way? Why not use the DHT client's timeout facilities, if that's what you need?

I don't feel it's a good idea to adapt the API to handle such unusual usage, unless you can explain a good reason why you have to do it that way.

gavin-norman-sociomantic:
Verbal discussion: we decided to adapt the app to use the availability tracker as intended (removing its own tracking of retry times). @vasileios-kyrillidis-sociomantic approves this PR :)

gavin-norman-sociomantic:
Unblocked following testing.

vasileios-kyrillidis-sociomantic (approved):

matthias-wende-sociomantic (approved):
:+1:

PR #185, opened 2018-10-18
Add turtle env method to wait until an expected number of listeners are active

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Blocked on testing in a test suite.

joseph-wakeling-sociomantic:
Needs an issue cross-ref ;-)  I'm going to try this out shortly, will get back to you with results.

joseph-wakeling-sociomantic:
Works for me!  Nice stuff :-)

I would anticipate that a common use-case would be to expect one listener for each of a specified list of channels, so could be nice to have that offered as an option.  But this is great already.  It allows me to get rid of the `delay` setting in test suite setup.

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

gavin-norman-sociomantic:
Good idea. Updated.

Would you be able to test the new, multi-channel method?

joseph-wakeling-sociomantic (approved):
LGTM!  This makes it possible to start integration tests with `delay = 0` and hence save a reasonable amount of time for each test-suite run.

joseph-wakeling-sociomantic:
I have tested with the app where we first observed a synchronization problem with startup and mirror requests, and it works beautifully.

gavin-norman-sociomantic:
Tested. Unblocked.

src/turtle/env/Dht.d
```diff
@@ -389,6 +389,62 @@ public class Dht : Node!(DhtNode, "dht")
...
+            check_interval = how often to poll for changes (seconds)
+
+        Throws:
+            TestException if timeout has been hit
+
+    ***************************************************************************/
+
+    public void expectListeners ( cstring[] channels, uint expected_listeners,
+        double timeout = 1.0, double check_interval = 0.05 )
+    {
```
matthias-wende-sociomantic, line 409: Why do you presume that the number of registered listeners is the same for all channels?

src/turtle/env/Dht.d
```diff
@@ -389,6 +389,62 @@ public class Dht : Node!(DhtNode, "dht")
...
+    }
+
+    /***************************************************************************
+
+        Waits until the specified number of listeners are registered with the
+        specified channels.
+
+        Params:
+            channels = DHT channels to lookup
+            expected_listeners = number of listeners to wait for
```
matthias-wende-sociomantic: This should be `number of listeners to wait for per channel`.

matthias-wende-sociomantic (commented):
This is great! It solves an issue I had half a year ago :D.
But I don't really understand why you can wait for listeners to get registered in different channels but assume that the number of listeners is always the same.

src/turtle/env/Dht.d
joseph-wakeling-sociomantic: It's a reasonable simplification for a common integration-test use case: you have a single app that is expected to listen to several different channels, hence the number of listeners can reasonably be assumed to match the total number of app nodes spawned.

If for some reason this assumption doesn't apply, you can use the single-channel overload on a channel-by-channel basis to deal with the custom numbers.

matthias-wende-sociomantic: So I presume the `expected_listeners` equals `1` usually, right?
I was mainly wondering about this generalisation that you can register more then once to a channel, but if you do then you have to register to all other channels `expected_listeners` times as well ;).

joseph-wakeling-sociomantic: Unless one is running integration tests with multiple app nodes, yes.

matthias-wende-sociomantic: My point here is, say we have a integration test scenario with more the one app registering to more then one channels. Then with this method would imply that all apps have to register to the same channels, which for me somehow contradicts this level of abstraction. What I would assume then is `expected_listeners` to be a map from `channel` → `nr. of expected_listeners for this channel`.

joseph-wakeling-sociomantic: @matthias-wende-sociomantic Do you have an actual scenario like that?

Note that nothing stops you using the single-channel overload to handle that case.  This is just a convenience overload to deal with a very common use-case.

joseph-wakeling-sociomantic: To be more precise, what I'd recommend for that use-case:

* use the array-based overload to handle multiple channels where the number of listeners is expected to be the same

* use the single-channel-name overload to handle channels expected to have a unique number of listeners

I'm not anti the idea of a map-based overload, but it's not really clear what benefit that would offer over the above suggestion, and in any case I think we should have an actual use-case before implementing something like that.  We do however _definitely_ have a use-case for the existing array-based overload with a common number of listeners.

matthias-wende-sociomantic: Alright, I see that the scenario having different apps running in a turtle environment is unlikely to happen any time soon, whereas having several nodes of one application (registering all to the same channels) is a viable test cast for some apps.

src/turtle/env/Dht.d
gavin-norman-sociomantic: Updated.

PR #197, opened 2019-02-05
Fix "race condition" in mirror helper

matthias-wende-sociomantic:
Assigning a request and changing the state are
assumed to be one atomic operation. However
assign calls other functions that might lead
into situations where the state is checked
although not yet changed.
Interchanging these commands solves this
problem in the given non threaded environments.


gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Please merge when the CI passes :)

PR #203, opened 2019-04-30
Stop using deprecated RecordBatch.DefaultMaxBatchSize

don-clugston-sociomantic:
RecordBatch.DefaultMaxBatchSize is marked as deprecated, yet is still used by dhtproto itself. Change the code to use the identically-defined RecordBatcher.DefaultMaxBatchSize instead.

This removes the annoying deprecation warning.

### DIFF
--- a/src/dhtproto/client/legacy/DhtConst.d
+++ b/src/dhtproto/client/legacy/DhtConst.d
@@ -66,7 +66,7 @@ static:
 
     ***************************************************************************/
 
-    public const RecordSizeLimit = RecordBatch.DefaultMaxBatchSize;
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic:
Slightly elaborated version: #204.

don-clugston-sociomantic:
Excellent! Your PR would be even better, if it would compile :)

matthias-wende-sociomantic:
Another nice thing is that we merged the (almost) same PR  in the neo branch last year :). https://github.com/sociomantic-tsunami/dhtproto/pull/160. (Exclude me in the we)

PR #209, opened 2019-05-24
Change storage engine getter to return via delegate

### LINKED ISSUE #144: Change node requests' storage engine getters to return values via a delegate
The storage implementation in the real DHT node does the following:
1. Calls the TokyoCabinet function to get a record's value. This `malloc`s a buffer and copies the record value into it.
2. Copies the value from the TC buffer into the request's buffer.
3. Frees TC's buffer.
4. (Probably copies the request's buffer into a send buffer somewhere. I forget exactly.)

There's at least one copy too many there. It's possible to refactor things to directly return the TC buffer and free it when finished (e.g. make the value accessible in the scope of a delegate), avoiding the extra copy.

For example, change this ([legacy Get request](https://github.com/sociomantic-tsunami/dhtproto/blob/v14.x.x/src/dhtproto/node/request/Get.d)):
```D
    // Get record for storage and send to client.
    final override protected void get ( cstring channel_name, cstring key )
    {
        this.writer.write(DhtConst.Status.E.Ok);
        this.writer.writeArray(this.getValue(channel_name, key));
    }

    // Get value from storage.
    abstract protected Const!(void)[] getValue ( cstring channel_name, cstring key );
```
to this:
```D
    // Get record for storage and send to client.
    final override protected void get ( cstring channel_name, cstring key )
    {
        this.writer.write(DhtConst.Status.E.Ok);
        this.getValue(channel_name, key,
            ( Const!(void)[] value )
            {
                this.writer.writeArray(value);
            }
        );
    }

    // Get value from storage.
    abstract protected void getValue ( cstring channel_name, cstring key,
        void delegate ( Const!(void)[] value ) value_dg );
```

See https://github.com/sociomantic-tsunami/dhtnode/issues/19.

tiyash-basu-sociomantic:
This improves performance by reducing by one the number of times the
value(s) gets copied before being returned.

The requests affected in this change are Get, GetChannelSize, GetChannels,
GetNumConnections, GetResponsibleRange, GetSize.

Fixes #144

### DIFF
--- a/src/dhtproto/node/request/Get.d
+++ b/src/dhtproto/node/request/Get.d
@@ -91,7 +91,12 @@ public abstract scope class Get : SingleKey
         cstring key )
     {
         this.writer.write(DhtConst.Status.E.Ok);
-        this.writer.writeArray(this.getValue(channel_name, key));
+        this.getValue(channel_name, key,
+            ( const(void)[] value )
+            {
+                this.writer.writeArray(value);
+            }
+        );
     }
 
     /***************************************************************************
@@ -102,11 +107,13 @@ public abstract scope class Get : SingleKey
         Params:
             channel_name = name of channel to query
             key = key of record to find
+            value_getter_dg = The delegate that is called with the value.
 
         Returns:
             value of queried record, empty array if not found
 
     ***************************************************************************/
 
-    abstract protected Const!(void)[] getValue ( cstring channel_name, cstring key );
+    abstract protected void getValue ( cstring channel_name, cstring key,
+        scope void delegate ( const(void)[] ) value_getter_dg );
 }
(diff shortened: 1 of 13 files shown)


matthias-wende-sociomantic (approved):

src/dhtproto/node/request/Get.d
```diff
@@ -102,11 +107,13 @@ public abstract scope class Get : SingleKey
...
+            value_getter_dg = The delegate that is called with the value.
 
         Returns:
             value of queried record, empty array if not found
 
     ***************************************************************************/
 
-    abstract protected Const!(void)[] getValue ( cstring channel_name, cstring key );
+    abstract protected void getValue ( cstring channel_name, cstring key,
+        void delegate ( const(void)[] ) value_getter_dg );
```
matthias-wende-sociomantic: Should be a scope delegate to avoid closure allocations.
Same for all delegates in this PR.

tiyash-basu-sociomantic: Fixed!

matthias-wende-sociomantic (approved):

PR #213, opened 2019-08-06
Fix missing 'import ocean.transition'

don-clugston-sociomantic:
This module uses `mstring` and friends, but doesn't import them properly.
(The problem shows up after D2 conversion, when you try to compile with dmd 2.087).


gavin-norman-sociomantic:
Thanks!

PR #214, opened 2019-08-07
Fix non-public imports to allow compilation with DMD 2.087

don-clugston-sociomantic:
This does not address the use of deprecated features,, which still generate dozens of warnings, but it does fix the errors.

Two groups of compilation errors are fixed:
* The swarm mixins need access to the Notification type, so the imports of the Requests need to be made public.

* The NeoSupport Internal mixin has an Internals struct. The imports inside it need to be transitive.

The same changes need to be made in dmqproto and dlsproto.

### DIFF
--- a/src/dhtproto/client/mixins/NeoSupport.d
+++ b/src/dhtproto/client/mixins/NeoSupport.d
@@ -103,16 +103,16 @@ template NeoSupport ( )
 
         private struct Internals
         {
-            import dhtproto.client.request.internal.GetHashRange;
-            import dhtproto.client.request.internal.Put;
-            import dhtproto.client.request.internal.Get;
-            import dhtproto.client.request.internal.Remove;
-            import dhtproto.client.request.internal.Mirror;
-            import dhtproto.client.request.internal.GetAll;
-            import dhtproto.client.request.internal.GetChannels;
-            import dhtproto.client.request.internal.Exists;
-            import dhtproto.client.request.internal.RemoveChannel;
-            import dhtproto.client.request.internal.Update;
+            public import dhtproto.client.request.internal.GetHashRange;
+            public import dhtproto.client.request.internal.Put;
+            public import dhtproto.client.request.internal.Get;
+            public import dhtproto.client.request.internal.Remove;
+            public import dhtproto.client.request.internal.Mirror;
+            public import dhtproto.client.request.internal.GetAll;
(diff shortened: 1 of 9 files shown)


matthias-wende-sociomantic:
LGTM. Can you please add a commit that bumps the DMD version in the travis file to DMD 2.087?
https://github.com/sociomantic-tsunami/dhtproto/blob/v15.x.x/.travis.yml#L43 and
https://github.com/sociomantic-tsunami/dhtproto/blob/v15.x.x/.travis.yml#L45

matthias-wende-sociomantic (approved):

matthias-wende-sociomantic:
Approved. I leave it for @gavin-norman-sociomantic to merge.

.travis.yml
```diff
@@ -40,9 +40,9 @@ jobs:
         - <<: *test-matrix
           env: DMD=2.078.3.s* F=devel
         - <<: *test-matrix
-          env: DMD=2.086.* F=production
+          env: DMD=2.087.* F=production
```
gavin-norman-sociomantic: Just one question: bumping the DMD version is to confirm that the bug fixes in the previous commits work, right?

gavin-norman-sociomantic: (If so, maybe just add that to the commit message.)

don-clugston-sociomantic: No, it's because Matthias told me to do it :)

gavin-norman-sociomantic: Do you know why he suggested that?

gavin-norman-sociomantic: I'm just thinking as this is a patch release we should minimise changes.

don-clugston-sociomantic: The conversation with him is higher up in this PR. If it would make it simpler to review/accept, I can remove that commit. Should I do that?

gavin-norman-sociomantic: Hmm, I don't see a conversation, just one comment (https://github.com/sociomantic-tsunami/dhtproto/pull/214#issuecomment-519106682). Let's leave it for now until Matthias can explain himself.

matthias-wende-sociomantic: > I'm just thinking as this is a patch release we should minimise changes.

Good point. Should be updated earliest in the next minor.
@don-clugston-sociomantic can you please remove the CI change ;). (Sorry for the unnecessary noise)

don-clugston-sociomantic: Now it is back to the original :)

PR #217, opened 2019-09-06
Fix GetChannels handleRequest method

matthias-wende-sociomantic:
Fixes a bug introduced in the following patch:
https://github.com/sociomantic-tsunami/dhtproto/pull/209/files#diff-be09e0578c38d9e48e66fc5cbdfa2041L69


don-clugston-sociomantic:
LGTM

ben-palmer-sociomantic:
Do you need to adjust the fakedht so the tests will pass?

matthias-wende-sociomantic:
Seems like this change of semantics was on purpose. Thanks travis for showing me!

PR #219, opened 2019-10-30
Replace `Const!` with `const`

jens-mueller-sociomantic:
This change replaces all usages of `Const!` with the D2 builtin `const`.
`Const!` was only used for transitioning to D2.
I also verified that neither `Immut!` not `Inout!` are used in this code
base.

Fixes sociomantic-tsunami/dhtproto#218.


alaksiej-stankievicx-dunnhumby (approved):

don-clugston-sociomantic (approved):
LGTM

PR #220, opened 2019-10-31
Add support for Ubuntu bionic

johannes-riecken-dunnhumby:


Build.mak
```diff
@@ -4,6 +4,12 @@ override LDFLAGS += -llzo2 -lebtree -lrt -lgcrypt -lgpg-error -lglib-2.0 -lpcre
 override DFLAGS  += -w -de
 DC:=dmd-transitional
 
+# Ubuntu bionic requires builds to use position independent code and
+# dmd-transitional does not set the flag -fPIC by default
```
Geod24, line 8: You guys should really add it to `/etc/dmd-transitional.conf` instead of adding it to ALL your projects.

daniel-zullo: Makes sense giving most of the projects still depends on dmd-transitional. Initially we tried to avoid any change to dmd-transitional as the goal at the end is to jump to newer versions of the dmd compiler (upstream).

Geod24: It's just a config change at the package level.
As I mentioned in my previous comment: If you can't run a program out of the box with your compiler because there's a mandatory flag missing, the package for your distribution is broken. This is akin to  having to specify Phobos (or druntime) include path for every build.

Issue #225, opened 2020-03-17
Drop support for v14.x.x


Geod24:
According to https://github.com/sociomantic-tsunami/dhtproto/releases/tag/v15.0.0 it's been 9 months since v15.0.0 has been released, so v14.x.x should be EOL.

Geod24:
Done

PR #228, opened 2020-03-24
Update README: Mention v15 and update dependencies

Geod24:
Ultimately we should also remove v14.x.x, see #225


don-clugston-sociomantic:
LGTM

PR #229, opened 2020-04-03
Fixes for dmd 2.092

don-clugston-sociomantic:
After this fix is made, if the submodules are updated, then dhtproto compiles cleanly with the beta of dmd 2.092, except for dip25 warnings from two lines in swarm.


alaksiej-stankievicx-dunnhumby (approved):
LGTM

PR #230, opened 2020-08-13
Client's NodeHashRanges: Use typeof instead of deprecated Trait

Geod24:
```
This is deprecated since v5.0.0 but since DMD didn't show deprecation messages on templates,
it was not noticed until we attempted to remove the module.
```

Intended to unblock https://github.com/sociomantic-tsunami/ocean/pull/823
CC @joseph-wakeling-frequenz

### DIFF
--- a/src/dhtproto/client/internal/NodeHashRanges.d
+++ b/src/dhtproto/client/internal/NodeHashRanges.d
@@ -264,7 +264,6 @@ public struct NodeHashRange
 
 private class NodeHashRangesBase
 {
-    import ocean.core.Traits : ReturnTypeOf;
     import swarm.neo.AddrPort;
     import swarm.neo.client.ConnectionSet;
     import swarm.util.Hash : isWithinNodeResponsibility;
@@ -275,7 +274,7 @@ private class NodeHashRangesBase
     private static ulong order_counter;
 
     /// Convenience alias for the comparison type of AddrPort.
-    private alias ReturnTypeOf!(AddrPort.cmp_id) Addr;
+    private alias typeof(AddrPort.init.cmp_id()) Addr;
 
     /// Map of IP address -> node hash range.
     private NodeHashRange[Addr] node_hash_ranges;


src/dhtproto/client/internal/NodeHashRanges.d
```diff
@@ -275,7 +274,7 @@ private class NodeHashRangesBase
     private static ulong order_counter;
 
     /// Convenience alias for the comparison type of AddrPort.
-    private alias ReturnTypeOf!(AddrPort.cmp_id) Addr;
+    private alias typeof(AddrPort.init.cmp_id()) Addr;
```
joseph-wakeling-frequenz, line 277: Would be more idiomatic perhaps just to swap the `ocean.core.Traits` import out for `ocean.meta.types.Function` (which defines the "new" `ReturnTypeOf`) ... ?

Geod24: Don in his review said he couldn't get it to work, hence this approach.
Personally I would have gone with a very different code here (because the return of `opCmp` is always `int`), so I just decided to stick with what the author intended.

joseph-wakeling-frequenz: Ah, I see on looking that they are not the same.

`ocean.core.Traits.ReturnTypeOf`:

```D
template ReturnTypeOf( Fn )
{
    static if (is(Fn Fptr : Fptr*) && is(Fptr == function))
    {
        // anything implicitly convertible to function pointer
        // this also handles new Typedef struct
        alias ReturnTypeOf!(Fptr) ReturnTypeOf;
    }
    else static if( is( Fn Ret == return ) )
        alias Ret ReturnTypeOf;
    else
        static assert( false, "Argument has no return type." );
}
```

`ocean.meta.types.Function.ReturnTypeOf`:

```D
public template ReturnTypeOf ( Callable... )
{
    static assert (Callable.length == 1);

    static if (is(typeof(Callable[0]) == function))
    {
        alias typeof({
                ParametersOf!(Callable[0]) args;
                return Callable[0](args);
            } ()) ReturnTypeOf;
    }
    else
    {
        alias typeof({
                ParametersOf!(Callable[0]) args;
                Callable[0] call;
                return call(args);
            } ()) ReturnTypeOf;
    }
}
```

It would probably be nicer to fix the bug in the new template (we should open an issue for this), but if you want to push forward with the new release(s) then I think it's fine to keep the current patch and fix the underlying bug later.  Can you note the reason in the commit message though?  i.e. that the new `ReturnTypeOf` is not a drop-in replacement and that we are working around a bug?

Geod24: Well, on the long run:
- We should probably use `opCmp` and use `int`, so the issue becomes moot;
- When it comes to traits, we *could* use Phobos - because the only reason we don't is allocation;

And I think those solution are more future proof.

Geod24:
Expanded the commit message.

joseph-wakeling-frequenz (approved):
LGTM, thanks!

src/dhtproto/client/internal/NodeHashRanges.d
joseph-wakeling-frequenz: > When it comes to traits, we could use Phobos - because the only reason we don't is allocation

Good point.  Let's open an issue to plan incorporating non-problematic phobos features like traits.

I suggest we target v7.x.x for first phobos usage, what do you think?  That should give us time to make good decisions (and experiments).

PR #234, opened 2020-08-25
Drop compatibility for anything less than v2.092, update deps to HEAD, fix deprecations

Geod24:
```
Since we want to drop support for dmd-transitional,
we need to settle on the minimum supported version of DMD.
Due to the various deprecations and some codeggen bugs in -O -inline,
only v2.092.0 and higher is able to compile our code correctly.

Additionally we now only test on bionic instead of xenial,
and reduced coverage to only the latest version of the compiler.
```

Note: Computer will say no until we fix deprecations in Swarm.


Geod24:
I'll wait for https://github.com/sociomantic-tsunami/ocean/pull/843 to be merged but this is (finally) green: https://github.com/Geod24/dhtproto/runs/2000700989

Geod24:
Looks like the time has finally come.

